PR #1, opened 2017-03-29
Fix backend user fixture

IchHabRecht:
As there isn't any components directory in the webroot, the path to the be_users.xml fixture needs to be relative to the FunctionalTestCase file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,30 @@
+########################
+# TYPO3 CMS
+# global ignore file
+########################
+#
+# Don't ignore .rej and .orig as we want to see/clean files after conflict resolution.
+#
+# For local exclude patterns please edit .git/info/exclude.
+#
+########################
+#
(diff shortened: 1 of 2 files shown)


psychomieze:
Hey,

this path problem along with all other fixture related path problems should already have been fixed with the final extraction patch - we introduced the possibility to use PACKAGES: syntax alongside EXT: syntax for now.

Can you check if that fixes your problem?

IchHabRecht:
Hi,

It finally seems to work by using PACKAGE: (not PACKAGES:) and exposing the path to the vendor directory. Two comments from my site:

- is it on purpose you removed the pages_language_overlay.xml as fixtures while still shipping pages.xml and sys_language.xml?
- why is it necessary to provide an environment variable for resolving the path to the vendor dir? This testing-framework is always part of the vendor dir and should be able to resolve the correct path by using __DIR__ shouldn't it?

PR #9, opened 2017-05-03
[BUGFIX] Core Bootstrap overrides local changes in Test Bootstrap

maddy2101:
By merging the two available arrays, the local changes take
precedence over the default ones. So the manual passage of
e.g. database credentials for functional tests is possible again.


lolli42:
that should be oboselete with eafaac680acc34e6288a50d90cecc1d02dfb6e66

PR #10, opened 2017-05-03
[TASK] Disable Caching for testing runs

maddy2101:
Together with the related core change to the Bootstrap it
is now possible to disable the extbase caches. To profit from this
possibility, we need to set the allowCaching parameter for the
Boostrap to false.

The related core change is https://review.typo3.org/#/c/52695


maddy2101:
there is an attempt to disable the extbase cache in FunctionalTestCase.php line 266. Still, it seems to be overridden again later on. Thats why I took the more intrusive approach. Feedback welcome.

psychomieze:
Is this one still needed? Or was this fixed by the recent config change?

psychomieze:
No feedback here anymore, closing now.

PR #12, opened 2017-05-03
[FEATURE] Initialize BackendRouter for FunctionalTests

maddy2101:
The UriBuilder will fail if backend routes are not loaded.


Classes/Core/Testbase.php
```diff
@@ -538,6 +538,7 @@ public function setUpBasicTypo3Bootstrap($instancePath)
             ->baseSetup()
             ->loadConfigurationAndInitialize(true)
             ->loadTypo3LoadedExtAndExtLocalconf(true)
+            ->initializeBackendRouter()
```
lolli42: I think it is better to call this in FunctionalTestCase.php at the two places directly before loadExtensionTables() is called.

PR #13, opened 2017-05-03
[FEATURE] Provide possibility to initialize sane defaults for features

maddy2101:
For example the very basic / non existing config for image
manipulation will now be set by educated guessing about
server environment.


lolli42:
I think that should be done in general somewhere in the bootstrap where instance LocalConfiguration is created and then merged with it.

psychomieze:
I'm not actually a fan of guessing configuration in a test scenario, as that makes the tests pretty volatile to running in different environments. Which problems are you trying to solve / Can you describe the goal of the patch?

PR #14, opened 2017-05-05
[TASK] Extract DataHandling Framework from ext:core

maddy2101:
The classes in ext:core/Test/Functional/DataHandling/Framework
are pretty useful for DataHandler tests. So it comes in handy
to have them available in testing-framework instead as part of
sysext:core.


maddy2101:
here is the related core change: https://review.typo3.org/52728

oliverklee:
As this is PHP-7-only, should we have strict types, type hints for basic types and return type hints in the two added classes?

lolli42:
In general fine with me, should have entries in classMap for the two classe for b/w compat, though.

Issue #29, opened 2017-08-30
Fatal error in pareser::process 


tmotyl:
if you pass data like
```
$iterator = ‌array (
  'alternative' => '',
  'combinedIdentifier' => '1:/user_upload/kasper-skarhoj1.jpg',
  'creationTime' => 1375080761,
  'description' => '',
  'extension' => 'jpg',
  'hashedIdentifier' => 'a58963e60a4fcc31d1a869184a2d5d929dd27c8a',
  'identifier' => '/user_upload/kasper-skarhoj1.jpg',
  'indexed' => true,
  'link' => '',
  'mimeType' => 'image/jpeg',
  'missing' => false,
  'modificationTime' => 1374139442,
  'name' => 'kasper-skarhoj1.jpg',
  'nameWithoutExtension' => 'kasper-skarhoj1',
  'publicUrl' => 'fileadmin/user_upload/kasper-skarhoj1.jpg',
  'sha1' => '05d8c6dda534a0b9e7023c3031e60e4b49c3da40',
  'size' => 39037,
  'title' => 'T3BOARD',
  'type' => 2,
  'uid' => 126,
);
and $identifier = "alternative";
```
then you'll get an fatal error
```
Fatal error: Uncaught TypeError: Argument 2 passed to TYPO3\TestingFramework\Core\Functional\Framework\Frontend\Parser::addRecord() must be of the type array, string given, called in vendor/typo3/testing-framework/Classes/Core/Functional/Framework/Frontend/Parser.php on line 67 and defined in /vendor/typo3/testing-framework/Classes/Core/Functional/Framework/Frontend/Parser.php on line <i>83
```

Situation like that can happen when rendering an object having FAL relation using jsonview

PR #30, opened 2017-08-30
Fix fatal error in Parser::process

tmotyl:
Fixes: https://github.com/TYPO3/testing-framework/issues/29

### DIFF
--- a/Classes/Core/Functional/Framework/Frontend/Parser.php
+++ b/Classes/Core/Functional/Framework/Frontend/Parser.php
@@ -61,6 +61,9 @@ class Parser implements \TYPO3\CMS\Core\SingletonInterface
     protected function process(array $iterator, array $path = [])
     {
         foreach ($iterator as $identifier => $properties) {
+            if (!is_array($properties)) {
+                continue;
+            }
             $this->addRecord($identifier, $properties);
             $this->addPath($identifier, $path);
             foreach ($properties as $propertyName => $propertyValue) {


Classes/Core/Functional/Framework/Frontend/Parser.php
```diff
@@ -61,6 +61,9 @@ public function parse(array $structure, array $path = [])
     protected function process(array $iterator, array $path = [])
     {
         foreach ($iterator as $identifier => $properties) {
+            if (!is_array($properties)) {
```
NamelessCoder: Should this case perhaps be caught further up the call stack? If $properties isn't an array that might indicate some other error that needs to be handled, rather than ignoring it here. Disclaimer: I don't know enough to about the functional test framework to say if that's the case.

tmotyl: Sure I can work around it, eg. by changing typo3/sysext/extbase/Tests/Functional/Fixtures/Extensions/blog_example/Classes/Controller/ContentController.php::getStructure.
But In my opinion this method is wrong, becuase it assumes that every second array contains only arrays in properties. The check for is_array is done only inside inner foreach loop. 

here is the full datastructure I'm passing inside parse function
See that originalResource doesn't have a wrapper like:
'sys_file_reference:126'

```
‌array (
  'tt_content:297' => 
  array (
    'header' => 'Regular Element #1',
    'image' => 
    array (
      'sys_file_reference:126' => 
      array (
        'originalResource' => 
        array (
          'alternative' => '',
          'combinedIdentifier' => '1:/user_upload/kasper-skarhoj1.jpg',
          'creationTime' => 1375080761,
          'description' => '',
          'extension' => 'jpg',
          'hashedIdentifier' => 'a58963e60a4fcc31d1a869184a2d5d929dd27c8a',
          'identifier' => '/user_upload/kasper-skarhoj1.jpg',
          'indexed' => true,
          'link' => '',
          'mimeType' => 'image/jpeg',
          'missing' => false,
          'modificationTime' => 1374139442,
          'name' => 'kasper-skarhoj1.jpg',
          'nameWithoutExtension' => 'kasper-skarhoj1',
          'publicUrl' => 'fileadmin/user_upload/kasper-skarhoj1.jpg',
          'sha1' => '05d8c6dda534a0b9e7023c3031e60e4b49c3da40',
          'size' => 39037,
          'title' => 'T3BOARD',
          'type' => 2,
          'uid' => 126,
        ),
        'pid' => 89,
        'uid' => 126,
      ),
    ),
    'pid' => '89',
    'uid' => '297',
  )
)
```

NamelessCoder: Someone else needs to chip in here, but I think this structure is expected (no identifier as wrapper) for 1:1 relations and probably should be handled rather than ignored.

To me this still looks like there might be more to the problem than just ignoring these 1:1 relations, such as correcting the fixtures or parsing that array into an actual record that then gets added.

lolli42: Well, for the sake of the pending core patch, I tend to accept that patch for now. This part is an internal testing related structure anyway, guess we're relatively free to adapt here if we change stuff later.

lolli42: Codewise, that change is basically a sanitation, too.

PR #33, opened 2017-09-17
[BUGFIX] Fix symlink creation on windows

psychomieze:
- To get correct paths change cwd
- Resolve from path to absolute path before creation

### DIFF
--- a/Classes/Core/Testbase.php
+++ b/Classes/Core/Testbase.php
@@ -235,8 +235,9 @@ class Testbase
             'typo3_src/typo3' => $instancePath . '/typo3',
             'typo3_src/index.php' => $instancePath . '/index.php',
         ];
+        chdir($instancePath);
         foreach ($linksToSet as $from => $to) {
-            $success = symlink($from, $to);
+            $success = symlink(realpath($from), $to);
             if (!$success) {
                 throw new Exception(
                     'Creating link failed: from ' . $from . ' to: ' . $to,


IchHabRecht:
Hi Susi,

Unfortunately '../../../../' still links to a wrong folder what relates in wrong symlinks for typo3 folder and index.php file.

psychomieze:
@IchHabRecht Could you tell me more about your setup? Mine is Windows 10 and Powershell or Ubuntu Bash, both correctly create the links on a clean install after the patch set.

IchHabRecht:
Hi Susi,

Here are some more information about my system/setup.

System: Windows 7
Shell: cmd (started as admin)
Path: D:\Version Control\Git\TYPO3\TYPO3.CMS\master
typo3_src path without patch: `18.09.2017  22:15    <SYMLINKD>     typo3_src [../../../../]`
-> error path not accessible
typo3_src with this patch: `18.09.2017  22:18    <SYMLINKD>     typo3_src [D:\Version Control]`

Hopes this helps to analyse the problem.

psychomieze:
Hmm, just had time to check this again and it's still working here, I don't know how else to test.

Can you verify whether chdir works on your system? The paths you wrote are correct, the instancePath + 4x ../ equals original root and is therefore correct.

lolli42:
Hmm. The patch looks good from a unix point of view. I'll merge now to unblock susi on other functional tasks we're working on. Nicole, if that still fails on your side: It would be cool if you could analyze a bit deeper on what is going wrong and maybe try to come up with a solution that fits all. That would be awesome.

IchHabRecht:
I don't work with Windows 7, so I cannot analyse the error anymore.

PR #62, opened 2018-04-06
[FEATURE] Backup and Restore Environment Object in Tests

maddy2101:
The Environment object is immutable for any TYPO3 instance, but
tests can initialize their own. In oder to keep tests side
effect free, after each of such tests the original Environment
has to be restored.
New functions to backup and restore this object ready to be used
in Test classes provide an easy solution to that task.

### DIFF
--- a/Classes/Core/Unit/UnitTestCase.php
+++ b/Classes/Core/Unit/UnitTestCase.php
@@ -61,6 +61,17 @@ abstract class UnitTestCase extends BaseTestCase
     private static $backupErrorReporting;
 
     /**
+     * The Environment object is used in TYPO3 to pass immutable settings
+     * like paths and system info around.
+     * It may be created in tests, but needs to be restored afterwards
+     * The array holds the original data to reset the Environment object
+     * after test run.
+     *
+     * @var array
+     */
+    private $backupedEnvironment = [];
+
+    /**
      * Set error reporting to trigger or suppress E_NOTICE
      */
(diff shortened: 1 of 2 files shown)


Classes/Core/Unit/UnitTestCase.php
```diff
@@ -60,6 +60,17 @@
...
+    /**
+     * The Environment object is used in TYPO3 to pass immutable settings
+     * like paths and system info around.
+     * It may be created in tests, but needs to be restored afterwards
+     * The array holds the original data to reset the Environment object
+     * after test run.
+     *
+     * @var array
+     */
+    private $backupedEnvironment = [];
```
oliverklee: This should be "backedUp", not "backuped".

Classes/Core/Unit/UnitTestCase.php
```diff
@@ -153,5 +172,41 @@ protected function tearDown()
...
+        if ($this->backupEnvironment === true) {
+            $this->restoreEnvironment();
+        }
+    }
+
+    /**
+     * before using Environment::initialize() in tests, backup the current data to be able to restore it afterwards
+     */
+    protected function backupEnvironment() {
+        $this->backupedEnvironment['context'] = TYPO3\CMS\Core\Core\Environment::getContext();
```
psychomieze: you only renamed the class var but still use the old one?

PR #66, opened 2018-05-06
[TASK] Simplify useExistingSession setup

maddy2101:
Almost all Cests include the fully loaded routine
right after the useExistingSession call. In order
to ease the setup for new Cests, include the routine
into useExistingSession


Resources/Core/Build/Configuration/Acceptance/Support/AcceptanceTester.php
```diff
@@ -53,5 +53,9 @@ public function useExistingSession()
 
         // reload the page to have a logged in backend
         $I->amOnPage('/typo3/index.php');
+        
+        // Ensure main content frame is fully loaded, otherwise there are load-race-conditions
+        $I->switchToIFrame('list_frame');
+        $I->waitForText('Web Content Management System');
```
lolli42: we should also add $I->switchToIFrame(); here, so the default is to be on the main iframe after default login.

PR #82, opened 2018-07-04
Refactored Test Framework

DavertMik:


Classes/Core/Acceptance/Helper/AbstractPageTree.php
```diff
@@ -1,5 +1,6 @@
 <?php
-namespace TYPO3\TestingFramework\Core\Acceptance\Page;
+declare(strict_types=1);
```
tmotyl: spaces around =

PR #103, opened 2018-10-03
[BUGFIX] Use correct path to autoload.php

IchHabRecht:
Using the subtree packages of TYPO3, the typo3 folder isn't a symlink
anymore. Thus the realpath cannot be used to resolve the path to
autoload. As the testing framework requires composer mode, the path
can simply be resolved from the current directory.

### DIFF
--- a/Classes/Core/Testbase.php
+++ b/Classes/Core/Testbase.php
@@ -609,7 +609,7 @@ class Testbase
         GeneralUtility::purgeInstances();
         GeneralUtility::resetApplicationContext();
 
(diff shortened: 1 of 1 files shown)


psychomieze:
Tested in project setups, core testing and for an extension.

psychomieze:
Made an error with testing this patch as it does _not_ work anymore for core. This is now causing a regression and will result in the core bamboo tests to fail. @IchHabRecht can you take a look, too? else I'd revert this tomorrow.

PR #116, opened 2019-01-07
[!!!][TASK] Remove constants and properly set up environment

bmack:
Environment API is now in place in favor of constants
* PATH_site
* PATH_thisScript
* TYPO3_OS

Setting up the constants is now removed, and Environment
initialization for Unit Tests is now working.


bmack:
won't work. too many broken / wrong tests in Core itself.

PR #120, opened 2019-01-11
[TASK] Drop definition of PATH_site

lolli42:


Resources/Core/Build/UnitTestsBootstrap.php
```diff
@@ -30,15 +30,34 @@
...
-    $testbase->createDirectory(PATH_site . 'uploads');
+
+    $requestType = \TYPO3\CMS\Core\Core\SystemEnvironmentBuilder::REQUESTTYPE_BE | \TYPO3\CMS\Core\Core\SystemEnvironmentBuilder::REQUESTTYPE_CLI;
+    \TYPO3\CMS\Core\Core\SystemEnvironmentBuilder::run(0, $requestType);
+
+    $testbase->createDirectory(\TYPO3\CMS\Core\Core\Environment::getPublicPath() . '/typo3conf/ext');
+    $testbase->createDirectory(\TYPO3\CMS\Core\Core\Environment::getPublicPath() . '/typo3temp/assets');
+    $testbase->createDirectory(\TYPO3\CMS\Core\Core\Environment::getPublicPath() . '/typo3temp/var/tests');
+    $testbase->createDirectory(\TYPO3\CMS\Core\Core\Environment::getPublicPath() . '/typo3temp/var/transient');
+    $testbase->createDirectory(\TYPO3\CMS\Core\Core\Environment::getPublicPath() . '/uploads');
```
bmack: you can get rid of this line btw.

Resources/Core/Build/UnitTestsBootstrap.php
```diff
@@ -30,15 +30,34 @@
...
-    $testbase->createDirectory(PATH_site . 'typo3conf/ext');
-    $testbase->createDirectory(PATH_site . 'typo3temp/assets');
-    $testbase->createDirectory(PATH_site . 'typo3temp/var/tests');
-    $testbase->createDirectory(PATH_site . 'typo3temp/var/transient');
-    $testbase->createDirectory(PATH_site . 'uploads');
+
+    $requestType = \TYPO3\CMS\Core\Core\SystemEnvironmentBuilder::REQUESTTYPE_BE | \TYPO3\CMS\Core\Core\SystemEnvironmentBuilder::REQUESTTYPE_CLI;
+    \TYPO3\CMS\Core\Core\SystemEnvironmentBuilder::run(0, $requestType);
+
+    $testbase->createDirectory(\TYPO3\CMS\Core\Core\Environment::getPublicPath() . '/typo3conf/ext');
```
bmack: could you use the respective usages within the environment API class instead?

bmack: also below....

lolli42: > could you use the respective usages within the environment API class instead?

will do with a next patch later. not sure if varPath is always correctly used / initialized in tests, yet.

Resources/Core/Build/UnitTestsBootstrap.php
lolli42: leads to a single test fail. let's do that separately

PR #122, opened 2019-02-09
[BUGFIX] Test for existing table before deleting from it

maddy2101:
According documentation, sqlite will create table sqlite_sequence
only if a table with AUTOINCREMENT exists in the schema.
This means for the functional tests of TYPO3, that maybe this
table does not exist at all. Trying to reset the sequence
will lead to an error then.
Checking for existence before manipulating prevents the error.


maddy2101:
not needed. Bug is related to pending core patch.

Issue #127, opened 2019-03-22
Allow all tables for functional test snapshotting


ohader:
https://bamboo.typo3.com/browse/CORE-GTN-FMS0PHP7201-712/test/case/98898673

```
TYPO3\CMS\Frontend\Tests\Functional\SiteHandling\EnhancerLinkGeneratorTest::localeModifierIsApplied with data set "localeModifier/enhancer:Extbase, lang:0" (array('Extbase', array(array('/{enhance_name}/{value}', 'Link::index', array())), 'testing', 'link', array(array('LocaleModifier', 'enhance', array(array('fr_FR', 'augmenter'))))), '&tx_testing_link[value]=100&t...=index', 0, 'https://acme.us/welcome/enhan...cHash=', array())
[...]

PR #128, opened 2019-03-23
[BUGFIX] Include all tables for data insertion

andreasfernandez:
This commit doesn't exclude tables without TCA anymore for data insertion.
This fixes an issue with mssql where inserting values into an identity
column fails due to missing TCA.

Resolves: #127


andreasfernandez:
I ran the test suite via `./Build/Scripts/runTests.sh -d mssql -s functional`:

Time: 57.12 minutes, Memory: 484.25 MB

OK, but incomplete, skipped, or risky tests!
Tests: 2756, Assertions: 17837, Skipped: 54, Incomplete: 7.

PR #131, opened 2019-04-26
[TASK] Add basic POST support to executeFrontendRequest.

sven-carstens:
* The http method verb is now taken from the request instead of being fixed to GET
* body content is passed along to the request

### DIFF
--- a/Classes/Core/Functional/Framework/Frontend/InternalRequest.php
+++ b/Classes/Core/Functional/Framework/Frontend/InternalRequest.php
@@ -63,8 +63,9 @@ class InternalRequest extends Request implements \JsonSerializable
      * @param string|null $uri URI for the request, if any.
      * @param string $method method to use, GET is default.
      * @param string|null $content body content to use, if any.
+     * @param array $headers Headers for the message, if any.
      */
-    public function __construct($uri = null, $method = 'GET', string $content = null) {
+    public function __construct($uri = null, $method = 'GET', string $content = null, array $headers = []) {
         if ($uri === null) {
             $uri = 'http://localhost/';
         }
@@ -73,7 +74,7 @@ class InternalRequest extends Request implements \JsonSerializable
             $body->write($content);
             $body->rewind();
         }
-        parent::__construct($uri, $method, $body);
+        parent::__construct($uri, $method, $body, $headers);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


sven-carstens:
* request headers are now passed along to the inside

maddy2101:
Can you give me some use case description? Why do you want this change? Probably a core patch that relies on this change, so please let me know in order to properly test the change.

sven-carstens:
I want to do some functional tests for a middleware.
The middleware not only supports GET but also POST requests.
Without this patch the HTTP Verb is fixed to "GET" abd there is no way to get anything else to the tested middleware.

With the patch you can set other methods and you can also set the request body.

Classes/Core/Functional/Framework/Frontend/InternalRequest.php
```diff
@@ -61,13 +61,20 @@ function (array $data) {
 
     /**
      * @param string|null $uri URI for the request, if any.
+     * @param string $method method to use, GET is default.
+     * @param string|null $content body content to use, if any.
+     * @param array $headers Headers for the message, if any.
      */
-    public function __construct($uri = null) {
+    public function __construct($uri = null, $method = 'GET', string $content = null, array $headers = []) {
```
bmack: how about we'll add an (optional) stream instead of a string?

sven-carstens: Whatever is not converted too many times firther down the road.

sven-carstens: Strangely moving the
```
        $body = new Stream('php://temp', 'wb+');
        if (null !== $content) {
            $body->write($content);
            $body->rewind();
        }
```

to the outer class and changing the type of the parameter broke everything. The body does not get passed along to the ServerRequestFactory

simonschaufi:
@sven-carstens @bmack can this PR be finished? I would love to use this feature!

Classes/Core/Functional/Framework/Frontend/InternalRequest.php
sven-carstens: > Strangely moving the
> 
> ```
>         $body = new Stream('php://temp', 'wb+');
>         if (null !== $content) {
>             $body->write($content);
>             $body->rewind();
>         }
> ```
> 
> to the outer class and changing the type of the parameter broke everything. The body does not get passed along to the ServerRequestFactory

Sorry, I give up on this.

PR #132, opened 2019-05-11
[TASK] Update phpunit to version 8

maddy2101:
composer require phpunit/phpunit:^8

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "issues": "https://github.com/TYPO3/testing-framework/issues"
   },
   "require": {
-    "phpunit/phpunit": "^8",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
     "issues": "https://github.com/TYPO3/testing-framework/issues"
   },
   "require": {
-    "phpunit/phpunit": "^7.5.6",
+    "phpunit/phpunit": "^8",
```
bmack: we still could do php 7.5 + 8?

bmack: aah, I meant "phpunit/phpunit:^7.5.7 || ^8" ?

maddy2101: True, I adjusted the PR. No reason to hard require v8.

PR #133, opened 2019-05-15
[TASK] Use correct cache name

maddy2101:


bmack:
Belongs to https://review.typo3.org/c/Packages/TYPO3.CMS/+/59661

PR #145, opened 2019-08-05
[WIP] Add basic POST support and add request headers to the internal …

sven-carstens:
…request.

This is related to https://github.com/TYPO3/testing-framework/pull/131 for TYPO3 V9 but for V10 instead.

### DIFF
--- a/Classes/Core/Functional/Framework/Frontend/InternalRequest.php
+++ b/Classes/Core/Functional/Framework/Frontend/InternalRequest.php
@@ -61,13 +61,20 @@ class InternalRequest extends Request implements \JsonSerializable
 
     /**
      * @param string|null $uri URI for the request, if any.
+     * @param string $method method to use, GET is default.
+     * @param string|null $content body content to use, if any.
+     * @param array $headers Headers for the message, if any.
      */
-    public function __construct($uri = null) {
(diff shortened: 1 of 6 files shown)


bmack:
seems legit. Can you update the PR. And can you elaborate why we need to create a new Application class?

sven-carstens:
The core AbstractApplication.php uses a static call to **\TYPO3\CMS\Core\Http\ServerRequestFactory::fromGlobals()** inside of
`final public function run(callable $execute = null)
    {
        try {
            $response = $this->handle(
                \TYPO3\CMS\Core\Http\ServerRequestFactory::fromGlobals()
            );
            if ($execute !== null) {
                call_user_func($execute);
            }
        } catch (ImmediateResponseException $exception) {
            $response = $exception->getResponse();
        }

        $this->sendResponse($response);
    }`

But inside of fromGlobals we need to read the body and the headers from the test call and not from the original php://input and the original headers.

Diff of the fromGlobals between the core ServerRequestFactory and the new derived type:
image
image

lolli42:
I think we can close this one for now: There have been two other recent merges which should allow most of what you're doing with the patch. Also, the new v11 way to do FE calls is executeFrontendSubRequest(), which suffers from less issues in this area.

PR #148, opened 2019-09-12
[TASK] Remove check for "pid=-1" on versioned records

bmack:
This is not needed anymore, as it is corresponding to the t3ver_oid>0 check which is included as well.

### DIFF
--- a/Classes/Core/Functional/Framework/DataHandling/ActionService.php
+++ b/Classes/Core/Functional/Framework/DataHandling/ActionService.php
@@ -487,9 +487,8 @@ class ActionService
 
         $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)
             ->getQueryBuilderForTable($tableName);
-        $queryBuilder->getRestrictions()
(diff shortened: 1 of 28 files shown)


Classes/Core/Functional/Framework/DataHandling/ActionService.php
```diff
@@ -491,10 +491,6 @@ protected function getVersionedId(string $tableName, $liveUid)
             ->select('uid')
             ->from($tableName)
             ->where(
-                $queryBuilder->expr()->eq(
-                    'pid',
-                    $queryBuilder->createNamedParameter(-1, \PDO::PARAM_INT)
```
ohader: Probably might select discarded version records... which should not be resolved as "live version"

ohader: Please adjust restrictions like this

```
$queryBuilder->getRestrictions()->removeAll()->add(GeneralUtility::makeInstance(DeletedRestriction::class));
```

PR #150, opened 2019-11-04
[TASK] Clean up code base

bmack:
- Use FQDN for use statements
- Fix some variable names and phpdoc info (typos mostly)
- Remove unused functionality
- Simplify code (e.g. notices via PHPunit directly)
- Use native PHP functionality over Core APIs when useful

### DIFF
--- a/Classes/Core/Functional/Framework/Frontend/Response.php
+++ b/Classes/Core/Functional/Framework/Frontend/Response.php
@@ -19,7 +19,6 @@ namespace TYPO3\TestingFramework\Core\Functional\Framework\Frontend;
  */
 class Response
 {
-    const STATUS_Success = 'success';
     const STATUS_Failure = 'failure';
 
     /**
--- a/Classes/Core/Functional/Framework/Frontend/ResponseContent.php
+++ b/Classes/Core/Functional/Framework/Frontend/ResponseContent.php
@@ -35,11 +35,6 @@ class ResponseContent
     /**
      * @var array
      */
-    protected $structurePaths;
-
-    /**
-     * @var array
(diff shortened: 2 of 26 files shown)


Classes/Core/Functional/Framework/Frontend/Response.php
```diff
@@ -19,7 +19,6 @@
  */
 class Response
 {
-    const STATUS_Success = 'success';
```
ohader: Not sure about changes in the `Response` layer... probably currently not used, but might be useful in future™

Classes/Core/Functional/Framework/Frontend/ResponseContent.php
```diff
@@ -108,14 +103,6 @@ function (string $sectionIdentifier) {
         );
     }
 
-    /**
-     * @return array
-     */
-    public function getScope(): array
```
ohader: Not sure about changes in the `Response` layer... probably currently not used, but might be useful in future™

ohader: Adjust restrictions like
```
$queryBuilder->getRestrictions()->removeAll()->add(GeneralUtility::makeInstance(DeletedRestriction::class));```

bmack: We can re-add later. You know... Git...

Classes/Core/Functional/Framework/Frontend/Response.php
bmack: We could/should use status codes instead, once we need it

PR #161, opened 2019-12-02
[!!!][TASK] Drop support for "callInaccessibleMethod" and "inject"

bmack:
The two helper functions in the "BaseTestCase" should
either be built manually by using the accessible mock
or by using closures to change the scope of the method.


bmack:
See https://review.typo3.org/c/Packages/TYPO3.CMS/+/62503

PR #167, opened 2020-02-01
[TASK] Allow use of Fluid 3.0

NamelessCoder:
Allow both Fluid ^2.5 and ^3 in tested systems.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
   "require": {
     "phpunit/phpunit": "^8.4",
     "mikey179/vfsstream": "~1.6.8",
-    "typo3fluid/fluid": "^2.5",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -27,7 +27,7 @@
   "require": {
     "phpunit/phpunit": "^8.4",
     "mikey179/vfsstream": "~1.6.8",
-    "typo3fluid/fluid": "^2.5",
+    "typo3fluid/fluid": "^2.5|^3",
```
bmack: according to the composer documentation it is "||" 

https://getcomposer.org/doc/articles/versions.md#version-range

bmack:
All good!

PR #177, opened 2020-04-19
[TASK] Check existence of div.module only

andreasfernandez:
Due to issue https://forge.typo3.org/issues/91127 the module "Help >
About" can't be the default anymore, thus the acceptance tests must not
rely on this.

As a replacement, we check whether `div.module` can be found, which
should be sufficient for the intended check.


andreasfernandez:
See tests results at https://bamboo.typo3.com/browse/CORE-GTC-2845 compared to https://bamboo.typo3.com/browse/CORE-GTC-2844

PR #186, opened 2020-07-20
[TASK] Use session snapshots to speed up tests with BE login

maddy2101:
BE login on each and every test adds a small amount of time to
each test setup. Codeception provides a possibility
to save a snapshot of such a session that can be reused, reducing
the set up time for tests.


maddy2101:
The measurable effect of the change is small (saves about half a minute for the FormEngine Tests on my local setup), but it will for sure not hurt. I got 10.4 minutes without and 9.8 minutes with the change.

PR #187, opened 2020-08-01
[TASK] Provide opportunity to write a site config for codeception tests

maddy2101:
In order to test multi language features, we need a more sophisticated
site configuration than the autogenerated one.

### DIFF
--- a/Classes/Core/Acceptance/Helper/AbstractSiteConfiguration.php
+++ b/Classes/Core/Acceptance/Helper/AbstractSiteConfiguration.php
@@ -32,6 +32,7 @@ abstract class AbstractSiteConfiguration
     protected const LANGUAGE_PRESETS = [
         'EN' => ['id' => 0, 'title' => 'English', 'locale' => 'en_US.UTF8'],
         'DK' => ['id' => 1, 'title' => 'Dansk', 'locale' => 'da_DK.UTF8'],
+        'DK' => ['id' => 2, 'title' => 'German', 'locale' => 'de_DE.UTF8'],
     ];
 
     /**
@@ -55,7 +56,8 @@ abstract class AbstractSiteConfiguration
         $configuration += [
             'langugages' => [
                 $this->buildDefaultLanguageConfiguration('EN', '/en/'),
-                $this->buildLanguageConfiguration('DK', '/dk/')
+                $this->buildLanguageConfiguration('DK', '/dk/'),
+                $this->buildLanguageConfiguration('DE', '/de/'),
             ],
         ];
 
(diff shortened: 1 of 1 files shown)


Classes/Core/Acceptance/Helper/AbstractSiteConfiguration.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @var AcceptanceTester
+     * 
+     * currently unused, but let's keep it for the time being. It will come in handy.
+     */
+    protected $tester;
+
+    public function adjustSiteConfiguration(): void
+    {
+        $sitesDir = __DIR__ . '/../../../../../../../typo3temp/var/tests/acceptance/typo3conf/sites';
```
bmack, line 46: Can we use the ORIGINAL_ROOT constant? seems "flaky" like that.

Classes/Core/Acceptance/Helper/AbstractSiteConfiguration.php
```diff
@@ -0,0 +1,66 @@
...
+        
+        $configuration = $this->buildSiteConfiguration(1, '/');
+        $configuration += [
+            'langugages' => [
+                $this->buildDefaultLanguageConfiguration('EN', '/en/'),
+                $this->buildLanguageConfiguration('DK', '/dk/')
+            ],
+        ];
+        
+       $siteConfiguration = new SiteConfiguration($sitesDir);
```
bmack, line 63: Any reason why you instantiate this class twice?

maddy2101: no, probably a logic error. Will correct.

Classes/Core/Acceptance/Helper/AbstractSiteConfiguration.php
maddy2101: I will try that, but I think you are right.

Classes/Core/Acceptance/Helper/AbstractSiteConfiguration.php
```diff
@@ -0,0 +1,65 @@
...
+{
+
+    use SiteBasedTestTrait;
+
+    /**
+     * @var array
+     */
+    protected const LANGUAGE_PRESETS = [
+        'EN' => ['id' => 0, 'title' => 'English', 'locale' => 'en_US.UTF8'],
+        'DK' => ['id' => 1, 'title' => 'Dansk', 'locale' => 'da_DK.UTF8'],
```
maddy2101, line 34: one more language here to stay in line with most of the functional tests

Issue #106, opened 2018-10-19
Add support to use the framework for functional tests in an own typo3 setup 


zdreicom:
If you have a TYPO3 website which comes with extensions you can not use the framework for functional.

I found to points wey it is not working:

1. typo3/sysext/core/bin/typo3

2. The way how to find the autoload.php is not really fancy it expected to have the typo3 folder on the project root.

markuspoerschke:
I just used the testing framework within a TYPO3 project. It was no problem at all. What are the problems you stumbled over when trying to use it?

Issue #144, opened 2019-07-28
load autoloader in FunktionalTest/Frontend Bootstrap broken


sHorst:
Hey, 

in file Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php on line 64 the framework try to load the composer autogenerated file. But this file is not on the same level as the index.php, but one folder up (as it should be out of the webroot) this means, the line should be corrected to

        `$directory = dirname(dirname(realpath($this->documentRoot . '/index.php')));`

The request object only gives this as error (path reduced for better readability)
  
>  [stderr] => PHP Warning:  require_once(.../.Build/Web/vendor/autoload.php): failed to open stream: No such file or directory in .../.Build/vendor/typo3/testing-framework/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php on line 64

Or did i setup my test setup wrong, i use the run_test.sh script and docker-compose.yml provided on the typo3 documentation.

sHorst:
Additional the Loading does not properly work when composer mode is on. So you should mark in the Documentation, that you need to set composer-mode to false in the composer.json of the extension.

```
	"extra": {
		"typo3/cms": {
...
			"composer-mode": false
		}
	}
```

sven-carstens:
Could you poste your whole composer.json setup?

For me the $directory from resolves to my PROJECT_DIR/.Build as intended and then finds its autoloader under PROJECT_DIR/.Build/vendor/autoload.php.

maritwho:
@sHorst Is there any update to this issue? I have the same problem and want to know, if there is a solution. thanks.

bmack:
Is this related to v4 or v5 branch?

bmack:
Just an update here: The testing framework is currently not working for composer mode, and no immediate plans are there to do so.

bmack:
(mostly due to time constraints)

sHorst:
can you add this to the documentation at least, will help a lot of people, who try to install it using composer :)

alexander-nitsche:
Hopefully with pull requests #189 and #190 the testing framework is working again with all scenarios of composer mode. At least my custom TYPO3 composer mode setup works fine with functional tests including frontend requests.

alexander-nitsche:
@sHorst , @sven-carstens : Can you confirm that your testing problems have gone with the latest testing-framework version 6.4.0?

mkempe:
@alexander-nitsche Thanks for your work, i can confirm that all problems are gone.

[...]

PR #189, opened 2020-08-06
[BUGFIX] Fix class loader path resolution in RequestBootstrap

alexander-nitsche:
Resolving the class loader path by using the public path is unreliable: The vendor folder does not have to be a child of the public folder but can have an arbitrary path. The most reliable and most used class loader path resolution in this extension is based on the fact that this extension itself resides within the vendor folder and thus has a fixed relation to the class loader file.

Resolves: #106, #144
Releases: master


bmack:
Seems legit to me, what do the others think?

Issue #188, opened 2020-08-06
Frontend Response: Class `TYPO3\JsonResponse\Encoder` not found


alexander-nitsche:
When running a functional test in my custom extension which includes a frontend request via
```
$this->executeFrontendRequest((new InternalRequest())->withPageId(2));
```
then the test breaks due to 
```
Frontend Response is erroneous: 
PHP Fatal error:  Uncaught Error: Class 'TYPO3\JsonResponse\Encoder' not found in /var/www/html/public/typo3/sysext/core/Classes/Utility/GeneralUtility.php:3423
```
When checking the according middleware class `TYPO3\JsonResponse\Encoder` in `/vendor/typo3/testing-framework/Resources/Core/Functional/Extensions/json_response/Classes/Encoder.php` it seems to be placed in an own extension _json\_response_ encapsulated within the extension _testing-framework_ without having its classes being loaded via `autoload.psr-4` in the `/vendor/typo3/testing-framework/composer.json`.

alexander-nitsche:
Having done some further investigations my fix seems to be brutal and not smart and i better remove it.

By design the classes of extension _json\_response_ get loaded only if `FunctionalTestCase->$frameworkExtensionsToLoad` includes that extension (which it does by default). Thus the classes of  _json\_response_ get loaded during the bootstrapping process of the functional test [here](https://github.com/TYPO3/testing-framework/blob/master/Classes/Core/Testbase.php#L583) and it can be used right away in the tests.

**But** as soon as the tests include a frontend request, the regarding response intercepting class `RequestBootstrap` seems to be running in a new context and has to bootstrap the whole application again including loading the classes [here](https://github.com/TYPO3/testing-framework/blob/master/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php#L58) and [there](https://github.com/TYPO3/testing-framework/blob/master/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php#L161). The class loading seems to forget the loading of the classes of  `FunctionalTestCase->$frameworkExtensionsToLoad` which is done in the `FunctionalTestCase` by invoking `ClassLoadingInformation::registerClassLoadingInformation();`. Thus simply adding this line to the bootstrapping process of class `RequestBootstrap` solves the issue:
```
/**
 * @return void
 */
public function executeAndOutput()
{
    ..
    try {
        chdir($_SERVER['DOCUMENT_ROOT']);
[...]

PR #190, opened 2020-08-06
[BUGFIX] Missing loading of json_response extension classes

alexander-nitsche:
Resolves: #188
Releases: master

### DIFF
--- a/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
+++ b/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
@@ -20,7 +20,6 @@ use TYPO3\CMS\Core\Core\Environment;
 use TYPO3\CMS\Core\Core\SystemEnvironmentBuilder;
 use TYPO3\CMS\Core\Utility\ArrayUtility;
 use TYPO3\CMS\Frontend\Http\Application;
-use TYPO3\TestingFramework\Core\Testbase;
 
 /**
  * Bootstrap for direct CLI Request
--- a/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
+++ b/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
@@ -15,6 +15,7 @@ namespace TYPO3\TestingFramework\Core\Functional\Framework\Frontend;
  */
 
 use TYPO3\CMS\Core\Core\Bootstrap;
+use TYPO3\CMS\Core\Core\ClassLoadingInformation;
 use TYPO3\CMS\Core\Core\SystemEnvironmentBuilder;
 use TYPO3\CMS\Core\Utility\ArrayUtility;
 use TYPO3\CMS\Frontend\Http\Application;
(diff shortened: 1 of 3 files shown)


Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
```diff
@@ -160,6 +161,7 @@ public function executeAndOutput()
             chdir($_SERVER['DOCUMENT_ROOT']);
             SystemEnvironmentBuilder::run(0, SystemEnvironmentBuilder::REQUESTTYPE_FE);
             $container = Bootstrap::init($this->classLoader);
+            ClassLoadingInformation::registerClassLoadingInformation();
```
bmack, line 164: Maybe we could add some notes why we would add this as Bootstrap::init() calls the same method as well (if in non-composer mode)

alexander-nitsche: It is aligned with the bootstrapping process of the FunctionalTestCase class [here](https://github.com/TYPO3/testing-framework/blob/master/Classes/Core/Testbase.php#L578) and [here](https://github.com/TYPO3/testing-framework/blob/master/Classes/Core/Testbase.php#L583). I would try to investigate why it was invented that way. Would you tell me, what you mean by non-composer setup: Is it the [Git + Composer Installation](https://get.typo3.org/version/10#accordion-git) (used for core development) or the [Wget / Curl / Download-Tar Installation](https://get.typo3.org/version/10#accordion-curl)?

alexander-nitsche: > non-composer setup = Git + Composer Installation

I tested the additional class loading with the Git + Composer installation as well and it showed that it was not required for that type of TYPO3 installation as the `Bootstrap::init()` one line above already loads those additional classes. Thus i added the condition `Environment::isComposerMode()` in order to handle that and to be in line with the `Bootstrap::init()` class load handling. 

To summarize: 
Functional tests in _composer mode_ and in _non-composer mode_ load the additional classes in `\TYPO3\TestingFramework\Core\Testbase::setUpBasicTypo3Bootstrap` via 
```
$this->dumpClassLoadingInformation();
``` 
– as some lines above in 
```
$container = Bootstrap::init($classLoader);
``` 
no composer autoload files are available yet and have to be dumped first.
In a frontend request, the regarding response intercepting class `RequestBootstrap` has to reload those additional classes which is done in `\TYPO3\TestingFramework\Core\Functional\Framework\Frontend\RequestBootstrap::executeAndOutput()` – for _non-composer mode_ in 
```
$container = Bootstrap::init($this->classLoader);
```
and for _composer mode_ in the code lines of this PR.

I also moved this loading of classes into a new method `Testbase::reloadClassLoadingInformation` to be aligned with method `Testbase::dumpClassLoadingInformation`.

Classes/Core/Testbase.php
```diff
@@ -598,6 +598,18 @@ public function dumpClassLoadingInformation()
...
     }
 
+    /**
+     * Reload class loading information if available
+     *
+     * @return void
+     */
+    public function reloadClassLoadingInformation()
+    {
+        if (ClassLoadingInformation::isClassLoadingInformationAvailable()) {
```
bmack, line 608: Hmm, this means, we reload ONLY if it exists, right? So it doesn't generate it "on the first run"?

bmack: We can also move this to the RequestBootstrap class directly.

alexander-nitsche: > Hmm, this means, we reload ONLY if it exists, right? So it doesn't generate it "on the first run"?

Yes, this function only loads classes which were dumped already. In RequestBootstrap we can be sure to have dumped classes already so we only need to load them again.

> We can also move this to the RequestBootstrap class directly.

Sure, i can move it back to RequestBootstrap directly.

alexander-nitsche: I moved it back to RequestBootstrap.

Issue #197, opened 2020-09-04
[BUG] Wrong middleware identifier in ext:json_response


ervaude:
While investigating https://review.typo3.org/c/Packages/TYPO3.CMS/+/65078 I found that the TF uses the nonexisting identifier `typo3/cms-frontend/site-resolver` for its middleware registration. With the linked core patch applied the MW is now not longer on the correct position. Also a middleware cannot be registered after the authentification MW and before the site resolver at the same time, so the configuration has to be changed.

lolli42:
resolved with the patch.

PR #198, opened 2020-09-04
[BUGFIX] Use correct identifier for middleware registration

ervaude:
Resovles: #197


ervaude:
Bamboo run with this patch is green: https://bamboo.typo3.com/browse/CORE-GTC-CDECC-5777/log

PR #200, opened 2020-09-10
[BUGFIX] Resolve new versioned records without t3ver_oid

bmack:
When removing the t3ver_state=1 (planned for TYPO3 v11)
placeholders which are only there to ease the pain
when publishing, the versioned record and the live record
are in fact the same, which should be covered by
the testing framework

### DIFF
--- a/Classes/Core/Functional/Framework/DataHandling/ActionService.php
+++ b/Classes/Core/Functional/Framework/DataHandling/ActionService.php
@@ -20,6 +20,7 @@ use TYPO3\CMS\Core\Database\ConnectionPool;
 use TYPO3\CMS\Core\Database\Query\Restriction\DeletedRestriction;
 use TYPO3\CMS\Core\DataHandling\DataHandler;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Core\Versioning\VersionState;
 use TYPO3\CMS\Workspaces\Service\WorkspaceService;
 use TYPO3\TestingFramework\Core\Exception;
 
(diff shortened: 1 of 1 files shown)


Classes/Core/Functional/Framework/DataHandling/ActionService.php
```diff
@@ -506,6 +507,41 @@ protected function getVersionedId(string $tableName, $liveUid)
         $row = $statement->fetch();
         if (!empty($row['uid'])) {
             $versionedId = (int)$row['uid'];
+        } else {
```
simonschaufi, line 510: change `return (int)$row['uid'];` in the line above and leave out the else.

simonschaufi (changes requested):

Classes/Core/Functional/Framework/DataHandling/ActionService.php
```diff
@@ -505,9 +506,43 @@ protected function getVersionedId(string $tableName, $liveUid)
...
+        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)
+            ->getQueryBuilderForTable($tableName);
+        $queryBuilder->getRestrictions()->removeAll()->add(GeneralUtility::makeInstance(DeletedRestriction::class));
+
+        $row = $queryBuilder
+            ->select('uid')
+            ->from($tableName)
+                ->where(
+                $queryBuilder->expr()->eq(
+                    'uid',
```
lolli42, line 523: this does not make sense: select('uid') -> where ('uid' = $liveuid) ???

Classes/Core/Functional/Framework/DataHandling/ActionService.php
```diff
@@ -505,9 +506,43 @@ protected function getVersionedId(string $tableName, $liveUid)
...
+                    't3ver_wsid',
+                    $queryBuilder->createNamedParameter($workspaceId, \PDO::PARAM_INT)
+                ),
+                $queryBuilder->expr()->eq(
+                    't3ver_oid',
+                    $queryBuilder->createNamedParameter(0, \PDO::PARAM_INT)
+                ),
+                $queryBuilder->expr()->eq(
+                    't3ver_state',
+                    $queryBuilder->createNamedParameter(VersionState::NEW_PLACEHOLDER_VERSION, \PDO::PARAM_INT)
```
lolli42, line 536: does not make sense either: t3ver_state=-1 records always have t3ver_oid != 0

lolli42 (changes requested):

lolli42 (approved):

lolli42:
Actually, this is fine with the pending removal placeholder change in v11

PR #203, opened 2020-10-02
Task/phpunit v9

maddy2101:


ohader:
→ duplicated from https://github.com/phpspec/prophecy-phpunit/blob/master/src/ProphecyTrait.php

PR #207, opened 2020-10-07
[BUGFIX] Use renamed class in phpunit v9

maddy2101:
class \Text_Template was available in phpunit/php-text-template
in version 1, but was renamed to \SebastianBergmann\Template\Template
in version 2.
So, using phpunit v9 and composer max conditions, the old class name
is not available anymore and causes failures in functional tests.

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -891,7 +891,12 @@ abstract class FunctionalTestCase extends BaseTestCase
         ];
 
         $vendorPath = (new Testbase())->getPackagesPath();
-        $template = new \Text_Template($vendorPath . '/typo3/testing-framework/Resources/Core/Functional/Fixtures/Frontend/request.tpl');
+        // @todo remove condition and if branch after phpunit v9 is minimum requirement
+        if (class_exists(\Text_Template::class)) {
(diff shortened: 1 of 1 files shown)


helhum (approved):

Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -891,7 +891,12 @@ protected function retrieveFrontendRequestResult(InternalRequest $request, Inter
         ];
 
         $vendorPath = (new Testbase())->getPackagesPath();
-        $template = new \Text_Template($vendorPath . '/typo3/testing-framework/Resources/Core/Functional/Fixtures/Frontend/request.tpl');
```
helhum, line 894: This solve the issue. For clarity and to avoid duplication I'd do it like this though:

```suggestion
        $templateClass = \Text_Template::class;
        if (!class_exists($templateClass)) {
        	$templateClass = \SebastianBergmann\Template\Template::class;
        }
        $template = new $templateClass($vendorPath . '/typo3/testing-framework/Resources/Core/Functional/Fixtures/Frontend/request.tpl');
```

helhum (changes requested):
Added a suggested code change. What do you think?

lolli42 (approved):

PR #216, opened 2020-11-26
[WIP][TASK] Set dependencies to TYPO3 v11 and phpunit v9

maddy2101:
this is preparation for testing-framework branch for TYPO3 v11 only. PHPUnit v9 will be required.


maddy2101:
change of plans

PR #217, opened 2020-11-29
Phpunit v9

maddy2101:


lolli42:
Dropping this PR for now: We'll branch a core v11-only (maybe v12-compat-too) of TF soon, which will then upgrade to latest phpunit as requirement. We'll do that with dedicated new patches.

PR #220, opened 2020-12-07
[TASK] Let tests that result in warnings fail

maddy2101:
This is for example the case for deprectation warnings.


lolli42:
Let us wait with this one until phpunit 9 is done.

lolli42:
Core now ships own .xml files since https://review.typo3.org/c/Packages/TYPO3.CMS/+/70239 to enable core specific settings.
Since this change *might* be at least slightly breaking for some usages, I'd say we should merge the patch when we branched-off next major version. This will be soon (as in: a couple of weeks).

PR #222, opened 2020-12-19
[TASK] Migrate to Bootstrap5

ohader:


lolli42:
How to continue here?

lolli42:
Checked: This patch would be ready: core v11 has an extending ModalDialog that already sets these selectors as advertised with the patch.
However, this patch would be breaking for v10.
We'll thus merge the patch as soon as TF is branched for core >= v11 and drop the override in core v11 along the way.
PR will be kept until branching.

bmack:
I think we should drop this here

PR #223, opened 2020-12-30
Backported #189 and #190 to `9` branch

mkempe:
Hey there,

first of all, many thanks for the `testing-framework` extension, its really great.

On the basis of the work from @alexander-nitsche i backported the changes to the `9` branch. All tests are running fine now with Composer. #144 should be fixed for TYPO3 9.x too.

@bmack Can you please look at the changes?

Thanks for your help.

### DIFF
--- a/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
+++ b/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
@@ -159,6 +159,10 @@ class RequestBootstrap
                 \TYPO3\CMS\Core\Core\SystemEnvironmentBuilder::REQUESTTYPE_FE
             );
             $container = \TYPO3\CMS\Core\Core\Bootstrap::init($this->classLoader);
+            if (\TYPO3\CMS\Core\Core\Environment::isComposerMode() &&
+                \TYPO3\CMS\Core\Core\ClassLoadingInformation::isClassLoadingInformationAvailable()) {
+                \TYPO3\CMS\Core\Core\ClassLoadingInformation::registerClassLoadingInformation();
+            }
             ArrayUtility::mergeRecursiveWithOverrule(
                 $GLOBALS,
                 $this->context->getGlobalSettings() ?? []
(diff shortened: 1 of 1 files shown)


lolli42:
Thanks for the backport Maik.
I have a personal issue with this patch, though: My brain ;) I have no good understanding of core v9 internals anymore and don't know which bits and pieces regarding file positioning and composer changed between v9, v10 and master. So I don't feel comfortable with merging this into v9 myself, and don't want to risk a regression in core-near package in a branch branch that is out-of-support in a couple of weeks.
Maybe anyone takes responsibility here and merges this into v9 compatible testing-framework branch, I'm a bit reluctant here.
I'll leave this PR open for a while to see if it is picked up.

alexander-nitsche:
Hi Maik, hi Christian,

my observations:

1) except for the changes of #189 and #190 the file Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php matches in both branches `master` and `9`, so there is a chance that nothing fundamentally has changed between `9` and `master` in that context. 
2) the change of #189 is the relative path from the file Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php to the parent directory of the testing-framework extension - the `vendor` folder. this path has definitely not changed between `9` and `master`.
3) the change of #190 has been tested by Maik and this change is only about composer mode explicitly which is not used in TYPO3 Core testing, as far as i am aware of. So this change would have no effect there even if it destroys the composer way of sending a request via the testing framework - which is currently broken anyway.

Maybe one or two additional test runs of this patch should be added here, to make this request be approved?

Greetings
Alex

mkempe:
Hi Christian and Alexander,

many thanks for taking your time to review the changes. I must admit that I have not really tested the changes without Composer, but with Composer is works fine, just like the master version. And as Alexander already said, the changes should not have a negative impact without Composer.

Here you can see the changes in action, Unit and functional tests for 9.x and 10.x: https://travis-ci.org/github/t3v/t3v_testing

PS: Please don't be surprised about `t3v_testing`, it is based on `testing-framework` and addional contains only a few helpers.

lolli42:
Ok. Convinced. Thx! Merged that now (btw: I was unable to push to the PR-branch to merge the branch with origin/9). I'll run it through core testing to verify it's green and will tag later.

lolli42:
Tagged 4.15.3

mkempe:
Many thanks ❤️.

PR #224, opened 2021-02-11
[TASK] Unset local variables in tearDown() in functional tests

lolli42:
Assigning the DI container as $this->container property in
functional tests introduced quite a memory hog when many
tests are executed.
Test instances are kept in phpunit until end-of-run, so
cleaning up huge objects assigned to test classes is
totally worth it.
A default tearDown() now unsets the container along with
other local properties.


bmack (commented):
👍

PR #229, opened 2021-04-22
[TASK] Include support for Lit v2 in JSUnit Bootstrap

bnf:
The javascript modules `@lit/reactive-element` and `lit`
will be available as of Lit v2.

Issue on Forge:
https://forge.typo3.org/issues/93965

Related change for TYPO3 core:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/68104

### DIFF
--- a/Resources/Core/Build/Configuration/JSUnit/Bootstrap.js
+++ b/Resources/Core/Build/Configuration/JSUnit/Bootstrap.js
@@ -24,9 +24,19 @@ var packages = [
 		main: 'lit-html'
 	},
 	{
+		name: '@lit/reactive-element',
+		location: '/base/typo3/sysext/core/Resources/Public/JavaScript/Contrib/@lit/reactive-element',
+		main: 'reactive-element'
+	},
+	{
 		name: 'lit-element',
 		location: '/base/typo3/sysext/core/Resources/Public/JavaScript/Contrib/lit-element',
 		main: 'lit-element'
+	},
+	{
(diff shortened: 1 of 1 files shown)


Resources/Core/Build/Configuration/JSUnit/Bootstrap.js
```diff
@@ -23,11 +23,21 @@ var packages = [
...
 	},
+	{
+		name: '@lit/reactive-element',
+		location: '/base/typo3/sysext/core/Resources/Public/JavaScript/Contrib/@lit/reactive-element',
+		main: 'reactive-element'
+	},
 	{
 		name: 'lit-element',
 		location: '/base/typo3/sysext/core/Resources/Public/JavaScript/Contrib/lit-element',
 		main: 'lit-element'
```
bnf, line 34: HEADS UP:
With `lit-element` v2 the `main` entry now actually is `index(.js)`.

We need to pick a decision here.
As we will not import from `lit-element` directly this currently doesn't matter for us.
For backwards compatibility for 3rd party extensions the core will use `index` here, but in testing framework we loose compatbility to 10.1/10.0 if we change this to `index`.

Therefore this pull requests stays with `lit-element` for now.

bmack: I think that's fine. we should keep this in mind when making the next major version of the testing framework.

PR #231, opened 2021-05-04
Explicitly check for valid input data

Crell:
Replaces #230 with a more draconian approach.  The root issue is invalid data that doesn't have a given field.  In that case, the input data is bad and should feel bad, so stop immediately.

This version uses a `ValueError`, because the input data is generally developer provided CSVs and thus a developer problem, not user problem.

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -747,6 +747,10 @@ abstract class FunctionalTestCase extends BaseTestCase
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


lolli42:
Looks as if ~200 functional tests are not clean, yet: https://gitlab.typo3.org/TYPO3/CI/cms/-/pipelines/11131  (that run uses your change, ignore the 'integrity' test fail, though - the functional fails are the relevant ones ...
Unsure yet, if we do a pre-patch in core first, or merge here and raise in core together with test bug fixes.

Issue #238, opened 2021-06-23
Allow to set header for internal request


DanielSiepmann:
It is possible to call something like this within a test: `$request = $request->withHeader('Content-Type', 'application/x-www-form-urlencoded');`.
This will break as the `AssignablePropertyTrait` will try to call `withHeaders()`.

```
Frontend Response is erroneous:

PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function TYPO3\CMS\Core\Http\Message::withHeader(), 1 passed and exactly 2 expected in /home/daniels/Projects/typo3/extensions/lbvgroup_jobs/.Build/web/typo3/sysext/core/Classes/Http/Message.php:202
Stack trace:

#0 [internal function]: TYPO3\CMS\Core\Http\Message->withHeader()
#1 vendor/typo3/testing-framework/Classes/Core/Functional/Framework/AssignablePropertyTrait.php(65): call_user_func()
#2 vendor/typo3/testing-framework/Classes/Core/Functional/Framework/Frontend/InternalRequest.php(50): TYPO3\TestingFramework\Core\Functional\Framework\Frontend\InternalRequest->with()
#3 vendor/typo3/testing-framework/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php(96): TYPO3\TestingFramework\Core\Functional\Framework\Frontend\InternalRequest::fromArray()
#4 .Build/web/typo3/sysext/core/Classes/Http/Message.php on line 202
```

A workaround could be to add the following to `InternalRequest`:
```
    public function withHeaders(array $headers): InternalRequest
    {
        $request = $this;
        foreach ($headers as $name => $value) {
            $request = $request->withHeader($name, $value);
        }

        return $request;
    }
```

DanielSiepmann:
Example usage:
```php

        $request = new InternalRequest($url);
        $request = $request->withHeader('Content-Type', 'application/x-www-form-urlencoded');
        $result = $this->executeFrontendRequest($request);
```

PR #239, opened 2021-06-23
Support withHeader() in InternalRequest

DanielSiepmann:
The request implements PSR-7 and therefore allows to call withHeader().
The AssignablePropertyTrait will be used to re build the request based
from serialization.
Headers are internally stored as array with name "headers".
The trait will therefore try to call withHeaders to add the stored
value.

This will break as no such method exists.
The message is added to stay compatible.

Resolves: #238


mbrodala (approved):

lolli42:
Pinged Daniel in private on this, asking for some details.

Issue #240, opened 2021-06-23
 Support POST Requests beside GET Requests


DanielSiepmann:
Extension Authors might need to test whether something behaves as expected when being called as POST.
E.g. controller or middleware.

DanielSiepmann:
Example usage:

```php
        $request = new InternalRequest($url);
        $request = $request->withMethod('POST');
        $request = $request->withHeader('Content-Type', 'application/x-www-form-urlencoded');
        $request = $request->withBody(/* insert post body */);
        $result = $this->executeFrontendRequest($request);
```

lolli42:
Setting this to done due to https://github.com/TYPO3/testing-framework/commit/d5cf8536ee837b9d0feec6f2dce132e782cebf01 being merged.

PR #241, opened 2021-06-23
Add basic support for none GET requests

DanielSiepmann:
Allow to use other methods beside GET from InternalRequest.
Use defined method.

Also support POST requests with Content-Type
'application/x-www-form-urlencoded'.

Relates: #240

### DIFF
--- a/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
+++ b/Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
@@ -104,6 +104,9 @@ class RequestBootstrap
 
         // Populating $_POST
         $_POST = [];
+        if ($this->request->hasHeader('Content-Type') && in_array('application/x-www-form-urlencoded', $this->request->getHeader('Content-Type'))) {
+            parse_str((string) $this->request->getBody(), $_POST);
+        }
         // Populating $_COOKIE
         $_COOKIE = [];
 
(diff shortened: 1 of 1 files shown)


Classes/Core/Functional/Framework/Frontend/RequestBootstrap.php
```diff
@@ -104,6 +104,9 @@ private function setGlobalVariables()
 
         // Populating $_POST
         $_POST = [];
+        if ($this->request->hasHeader('Content-Type') && in_array('application/x-www-form-urlencoded', $this->request->getHeader('Content-Type'))) {
```
mbrodala, line 107: What about `multipart/form-data`?

DanielSiepmann: Should also be added. I've currently added what I needed in my concrete use case.
Would need to check how that needs to be handled / parsed.

Should we add it within this PR, or would it be ok to not add it yet, but once someone needs that feature?

mbrodala: I'd vote for adding this separately since I'm also not 100% sure what further implications this has.

lolli42:
This only affects (deprecated) v10 way using executeFrontendRequest(), right?
We could add this, but what about v11 way using executeFrontendSubRequest()? Does it 'just work' there?

lolli42:
General note on InternalRequest:
I think it would be beneficial if InternalRequest would decorote Request (or ServerRequest). This would void the need to transfer the InternalRequest into a ServerRequest within TF sub request handling.
The approach would be similar to what we recently did with extbase Request, which is a decorator of ServerRequest, too. Maybe the TF Request object could do the same.
I think we should have a look at this construct when we finally branch-off a core v11-only version of TF which drops the non-sub-request methods.

lolli42:
Ok, when doing the new executeFrontendSubRequest(), it should basically 'just work' as far as I can see. At least if the system under test works with the $reqest object. If not, this should be solved with a core or extension patch anyways.
I'll merge that patch now.

PR #244, opened 2021-07-14
[BUGFIX] Provide minimal serverParams for Frontend functional subrequest

### LINKED ISSUE #246: normalizedParams are not setup correctly using v11 subrequest for frontend request testing
Provide minimal serverParams array for core v11 compatible frontend subrequest, so the normalized param attributes are set through corresponding frontend middleware, which is needed for some functional tests, ex. PageTypeSuffixDecorator tests.

See https://forge.typo3.org/issues/94537

PulLReuqest: #244

sbuerk:
Provide minimal serverParams array for core v11 compatible frontend subrequest, so the normalized param attributes are set through corresponding frontend middleware, which is needed for some functional tests, ex. PageTypeSuffixDecorator tests.

See https://forge.typo3.org/issues/94537

Resolves #246

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -971,11 +971,21 @@ abstract class FunctionalTestCase extends BaseTestCase
         $uriString = (string)$uri;
         $uri = new Uri($uriString);
 
+        // build minimal serverParams for normalizedparamsAttribute initialzation
+        $serverParams = [
+            'SCRIPT_NAME'           => $_SERVER['SCRIPT_NAME'],
+            'HTTP_HOST'             => $_SERVER['HTTP_HOST'],
+            'SERVER_NAME'           => $_SERVER['SERVER_NAME'],
+            'HTTPS'                 => $uri->getScheme() === 'https' ? 'on' : 'off',
+            'REMOTE_ADDR'           => '127.0.0.1',
+        ];
+
         $serverRequest = new ServerRequest(
             $uri,
             $request->getMethod(),
             'php://input',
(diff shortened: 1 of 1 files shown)


lolli42:
Thanks for the patch.
Yeah. It's very possible some _SERVER details are not properly set up for sub requests, yet and NormalizeParams is then not initialized correctly. I roughly remember that I stumbled upon something, too, but didn't investigate in detail, since I had enough to do with other details in this area, when I created the sub request stuff
It's also possible some details of Environment are not fully correct.
This patch needs a good review and a verification core tests are still green (or need adaption) before merge. Not fully sure on when I'll find time for this, if nobody else steps in.

sbuerk:
Thanks @lolli42 .

Yes I assumed that. I saw you added the frontend subrequest as a first thing, I thought this went under.

I had running locally the tests (unit,functional) before and after this "fix", and all was green (core tests that was on master).

For my "core patch" it was different. Before green, with theses changes:

lolli42:
Thanks for your elaboration, Stefan!
v10 and master currently use the same TF version (we'll branch-off soon, though). Sub request handling only works with v11 (this is actually a HUGE achievement in v11 core!). v9 can be ignored.
I hope to find some time for an in-depth review soon and your comment helps to get more confidence on this patch.

lolli42:
Verified the patch. Looks good and is the right thing to do. Note the GLOBALS['_SERVER'] assignments above will fall at some point when GU:getIndPEnv() is finally gone (we're working on this, but doesn't look as if it will be finished with v11).

lolli42:
Tagged 6.9.0 and raised in core master. Go ahead with your core patch, Stefan!

PR #249, opened 2021-07-19
[TASK] Add GitHub templates

alexander-nitsche:


.editorconfig
```diff
@@ -4,11 +4,16 @@
...
 charset = utf-8
 end_of_line = lf
 indent_style = space
-indent_size = 3
 insert_final_newline = true
 max_line_length = 80
 trim_trailing_whitespace = true
+
+[*.rst]
+indent_size = 3
```
lolli42, line 16: Pesonal itch: I *hate* the indent_size =3 in core. It is completely stupid and makes .rst file editing way harder than it should be, especially for code blocks. Let's set to 4 ...

alexander-nitsche: I would keep it at 3 spaces, as is in Core and Documentation _for reasons_. It is not completely stupid as it let look the lists a bit nicer. With indentation 3 it is

-  List item title
   List item content

with indentation 4 it is 

-   List item title
    List item content

PR #262, opened 2021-08-21
[TASK] Only truncate MySQL tables that have been touched

oliverklee:
"Touched" means "the auto increment has been changed (if there is
one)" or "there are some records in it".

This massively speeds up functional tests (as the table truncation
usually is the slowest part of the test setup).

### DIFF
--- a/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseAccessor.php
+++ b/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseAccessor.php
@@ -113,7 +113,7 @@ class DatabaseAccessor
             );
         }
 
-        $this->connection->truncate($tableName);
+        $this->truncate($tableName);
 
         $columnNames = array_keys($columns);
         foreach ($items as $item) {
@@ -122,6 +122,17 @@ class DatabaseAccessor
                 array_combine($columnNames, $item),
                 $columns
             );
+        }
+    }
+
+    private function truncate(string $tableName): void
+    {
+        $databaseName = $this->connection->getDatabase();
+        $query = "SHOW TABLE STATUS FROM $databaseName LIKE '$tableName'";
+        $tableData = $this->connection->executeQuery($query)->fetchAssociative();
+        $isChanged = ((int) $tableData['Auto_increment']) > 1 || ((int) $tableData['Rows']) > 0;
+        if ($isChanged) {
+            $this->connection->truncate($tableName);
         }
     }
 
--- a/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseAccessor.php
+++ b/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseAccessor.php
@@ -16,7 +16,6 @@ namespace TYPO3\TestingFramework\Core\Functional\Framework\DataHandling\Snapshot
  */
 
 use Doctrine\DBAL\FetchMode;
-use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\DBAL\Query\QueryBuilder as DoctrineQueryBuilder;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Connection;
@@ -114,7 +113,7 @@ class DatabaseAccessor
             );
         }
 
-        $this->truncate($tableName);
+        $this->connection->truncate($tableName);
 
         $columnNames = array_keys($columns);
         foreach ($items as $item) {
@@ -123,41 +122,6 @@ class DatabaseAccessor
                 array_combine($columnNames, $item),
                 $columns
             );
-        }
-    }
-
-    private function truncate(string $tableName): void
-    {
-        $platform = $this->connection->getDatabasePlatform();
-        if ($platform instanceof MySqlPlatform) {
-            $this->truncateForMySql($tableName);
-        } else {
-            // @todo: Optimize truncation for other platforms, too.
-            $this->connection->truncate($tableName);
-        }
-    }
-
-    /**
-     * Truncates a table for MySQL databases in an optimized way.
-     *
-     * This method tries to avoid the (expensive) truncate if possible:
-     * - If the table has an auto-increment value (which usually is the `uid` column`) and that value has changed,
-     *   this method will truncate the table.
-     * - If the table does not have an auto-increment value, but it has at least one row (where the exact number does
-     *   not matter), this method will truncate the table.
-     * - Otherwise, this method will skip the truncate. (For tables without an auto-increment value, this means that
-     *   the table either has not been touched at all beforehand, or that all records have already been deleted).
-     */
-    private function truncateForMySql(string $tableName): void
-    {
-        $databaseName = $this->connection->getDatabase();
-        $query = "SHOW TABLE STATUS FROM $databaseName LIKE '$tableName'";
-        $tableData = $this->connection->executeQuery($query)->fetchAssociative();
-        $hasChangedAutoIncrement = ((int)$tableData['Auto_increment']) > 1;
-        $hasAtLeastOneRow = ((int)$tableData['Rows']) > 0;
-        $isChanged = $hasChangedAutoIncrement || $hasAtLeastOneRow;
-        if ($isChanged) {
-            $this->connection->truncate($tableName);
         }
     }
 
--- a/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseAccessor.php
+++ b/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseAccessor.php
@@ -130,19 +130,33 @@ class DatabaseAccessor
     {
         $platform = $this->connection->getDatabasePlatform();
         if ($platform instanceof MySqlPlatform) {
-            // Optimized truncation for mysql:
-            // * Tables with auto increment (typically uid) > 0 - a row has been inserted here, but may have been deleted
-            // * And tables that have rows
-            $databaseName = $this->connection->getDatabase();
-            $query = "SHOW TABLE STATUS FROM $databaseName LIKE '$tableName'";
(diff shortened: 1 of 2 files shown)


sbuerk:
Mhh .. not tested, but some questions what would came up to my mind:

1. What if one or more records in table was deleted ?
2. What if one or more recoreds where updated ?
3. importTable is doing simple inserts after truncating .. so if not truncated, insert would (should) fail for existing primary key records

oliverklee:
@sbuerk Thanks, good questions! The code already covers all these cases, but (q.e.d.) fails to communicate that it does and why it does it. :-)

Instead of explaining the details here, I'm going to take your questions as a reminder that I need to rework the code so that it answers these questions to the reader.

sbuerk:
@oliverklee mhh .. as I said, not tested. Just "peeked" into the commit / changes itself.

But anway -- the biggest thing is, that changes are not compativble with sqlite testing.

```
$ cd /tmp && rm -rf /tmp/t3-core && git clone git://github.com/typo3/typo3 /tmp/t3-core && cd /tmp/t3-core
$ git fetch https://review.typo3.org/Packages/TYPO3.CMS refs/changes/40/70640/9 && git cherry-pick FETCH_HEAD
$ Build/Scripts/runTests.sh -s composerInstall
$ rm -rf vendor/typo3/testing-framework && git clone git@github.com:oliverklee/typo3-testing-framework.git vendor/typo3/testing-framework
$ cd vendor/typo3/testing-framework && git checkout task/faster-truncate && cd ../../..
$ Build/Scripts/runTests.sh -s composerInstall
$ Build/Scripts/runTests.sh -s functional -d sqlite typo3/sysext/frontend/Tests/Functional/SiteHandling/SiteRequestTest.php
```

Example exeption:
```
....

91) TYPO3\CMS\Frontend\Tests\Functional\SiteHandling\SiteRequestTest::pageIsRenderedWithValidCacheHash with data set "https://website.local/?&cHash=f42b850e435f0cedd366f5db749fc1af&id=1100&testing[value]=1" ('https://website.local/?&cHash...lue]=1')
Doctrine\DBAL\Exception\SyntaxErrorException: An exception occurred while executing 'SHOW TABLE STATUS FROM /var/www/work/typo3-play/master/src/typo3temp/var/tests/functional-0f6f00f/test.sqlite':

SQLSTATE[HY000]: General error: 1 near "SHOW": syntax error

/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractSQLiteDriver.php:75
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:182
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:159
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:2212
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:1299
/var/www/work/typo3-play/master/src/vendor/typo3/testing-framework/Classes/Core/Testbase.php:617
/var/www/work/typo3-play/master/src/vendor/typo3/testing-framework/Classes/Core/Functional/FunctionalTestCase.php:292
/var/www/work/typo3-play/master/src/typo3/sysext/frontend/Tests/Functional/SiteHandling/SiteRequestTest.php:52

Caused by
Doctrine\DBAL\Driver\PDO\Exception: SQLSTATE[HY000]: General error: 1 near "SHOW": syntax error

/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDO/Exception.php:18
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:134
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php:36
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:1295
/var/www/work/typo3-play/master/src/vendor/typo3/testing-framework/Classes/Core/Testbase.php:617
/var/www/work/typo3-play/master/src/vendor/typo3/testing-framework/Classes/Core/Functional/FunctionalTestCase.php:292
/var/www/work/typo3-play/master/src/typo3/sysext/frontend/Tests/Functional/SiteHandling/SiteRequestTest.php:52

Caused by
PDOException: SQLSTATE[HY000]: General error: 1 near "SHOW": syntax error

/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:132
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php:36
/var/www/work/typo3-play/master/src/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:1295
/var/www/work/typo3-play/master/src/vendor/typo3/testing-framework/Classes/Core/Testbase.php:617
/var/www/work/typo3-play/master/src/vendor/typo3/testing-framework/Classes/Core/Functional/FunctionalTestCase.php:292
/var/www/work/typo3-play/master/src/typo3/sysext/frontend/Tests/Functional/SiteHandling/SiteRequestTest.php:52
```

Not sure about postgres

oliverklee:
Yes, it indeed breaks both SQLite as well as PostgreSQL: https://review.typo3.org/c/Packages/TYPO3.CMS/+/70715

I'll need to add specific cases for these RDBMs (or fall back to using truncate for all tables for RDBMs that this change currently does not support yet).

I do welcome help on this!

sbuerk:
> @sbuerk Thanks, good questions! The code already covers all these cases, but (q.e.d.) fails to communicate that it does and why it does it. :-)
> 
> Instead of explaining the details here, I'm going to take your questions as a reminder that I need to rework the code so that it answers these questions to the reader.

Sorry to say, but I do not see where the code should cover the tables with just "updates" records (without new/or deleted records). Maybe you can explain in short how the code is covering this ?

DatabaseAccessor, which is also used for Snaptshots create/restore ... if auto_increment value is not changed, it will not truncate that table.  DatabaseAccessor->import() calls the truncate(), and simple loop through the data and doing "inserts".

Deleted data would be restored, but rows still existing will not be restored to orignal state, because the "insert()" calls in the import() method would fail (for tables with unique primary key / autoincrement tables).

In general I can understand why you are trying to get that faster ... but lean back and thinking a moment about it, to really check if you need to trunacate a table for all circumstances would take at leas the same time (you will have to check each data row if it is still the same), and having each row checks if already exists and update instead of insert and so on. So you would not gain anything speed wise.

Maybe I oversee where the code should cover this "already" ... but from execution order it would not. Maybe currently it works because there are no tests which only updates or delete something without adding something, and so no exceptions are thrown .. maybe a add dummy functional testcases to your core patch which special construct these cases would came up with this.

Not sure how to accomplisch database storage aware code .. I would suggest to rethinkg the notet states. Just points from myside. 

Another thing is, mysql information schema rows is not guaranteeed accurate. It is inaccurate. For simple > 0 checks it could be okay, but must not. For exact row count it is absolutly unusable / inaccurate due inno db issues.

oliverklee:
@sbuerk Okay, then I'll try to explain my approach.

The point of the changed methods is to either reset* all tables or a single table.
(* reset = drop all records and set the auto increment value to the initial value (if there is any))

We need to differentiate between two kinds of tables here:
1. Those with an auto increment value. If any record had been created, the auto increment value gets increased. Deleting the records does not reset the auto increment value. So checking if the auto increment value has changed is enough to determine whether the table needs to be really truncated or is untouched (and hence does not need truncation).
2. Those without an auto increment value. Here, we can check whether there is at least one record present. (The exact number does not matter - we only need to know whether it is zero or non-zero. This should cover your point concerning the information schema being inaccurate.) If there are any records in it, we need to truncate. And if there are no records in it (either because there have not been any to begin with or they have all been deleted), we can skip the truncation.

The function that is expected to truncate all tables loops over all tables (not over all records) and checks this:
1. if auto increment present and changed -> truncate
2. if no auto increment present: number of records non-zero -> truncate

Does this explanation help you understand my approach?

This change works with all existing core tests for MySQL. As you can see at https://gitlab.typo3.org/typo3/CI/cms/-/pipelines/15378, all MySQL-based tests pass. (The SQLite- and PostgreSQL-based tests fail because the SQL syntax for getting the DB schema needs to be different.)

Concerning the performance: Please let's look at the numbers instead of speculating.

I have done basically the same change for the nimut testing framework. There, the functional tests for my oelib extension on PHP 7.4 and TYPO3 9LTS go down from [5:10 mins](https://github.com/oliverklee/ext-oelib/runs/3389924446?check_suite_focus=true#logs) to [1:26 mins](https://github.com/oliverklee/ext-oelib/runs/3389926362?check_suite_focus=true).

For the core master tests, the first batch of MySQL-based functional tests on PHP 8.0 go down from (7:23 mins)[https://gitlab.typo3.org/typo3/CI/cms/-/jobs/413178] to (1:49 mins)[https://gitlab.typo3.org/typo3/CI/cms/-/jobs/411689] (for PHPUnit only, without the container setup, Composer install etc.)

lolli42:
Oliver. Your approach is great and the performance improvement is totally worth it!

sqlite needs a different handling: There is a way to look for tables that had 'sequences' changed, but there is no way to see if tables have rows, without firing a count() to single tables, when a table has no auto increment. But I'd have a different idea for sqlite that should be even more efficient.

postgres can probably do a similar strategy as mysql, it just needs some fiddling I suppose.

For now I changed the patch to "optimize truncation" only when running mysql/maria. This should do and we can merge this PR. The other dbms approaches need some refactoring before and can be done with dedicated patches ... I'll see what I can do.

lolli42 (approved):

PR #264, opened 2021-08-28
[TASK] Only truncate touched MySQL tables using real row counts

sbuerk:
This is a follow up to mitigate reliable issues for mysql introduced by #262 .

Using information schema for row counts is not reliable enough for innodb tables, and for mysql 8.0 it fail hard. 

Using subquery as a joinable table building information if tables has at least one row mitigates this issue.

### DIFF
--- a/Classes/Core/Testbase.php
+++ b/Classes/Core/Testbase.php
@@ -659,17 +659,49 @@ class Testbase
             ->getConnectionByName(ConnectionPool::DEFAULT_CONNECTION_NAME);
 
         $databaseName = $connection->getDatabase();
-        $query = "SHOW TABLE STATUS FROM $databaseName";
+        $tableNames = $connection->getSchemaManager()->listTableNames();
+
+        // no tables to process;
+        if (!$tableNames) {
+            return;
+        }
+
+        // build subselect to get joinable table with information if table has at least one row.
+        // This is needed because information_schema.table_rows is not reliable enough for innodb engine.
+        // see https://dev.mysql.com/doc/mysql-infoschema-excerpt/5.7/en/information-schema-tables-table.html TABLE_ROWS
+        $fromTableUnionSubSelectQuery = [];
+        foreach($tableNames as $tableName) {
+            $fromTableUnionSubSelectQuery[] = sprintf(
+                ' select %s as table_name, exists(select * from %s limit 1) as has_rows',
+                $connection->quote($tableName),
+                $connection->quoteIdentifier($tableName)
+            );
+        }
+        $fromTableUnionSubSelectQuery = implode(' UNION ', $fromTableUnionSubSelectQuery);
+
+        $query = sprintf(
+            "
+                select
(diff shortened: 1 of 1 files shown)


Classes/Core/Testbase.php
```diff
@@ -657,19 +653,47 @@ private function truncateAllTablesForMysql(): void
...
+            // No tables to process
+            return;
+        }
+
+        // Build a sub select to get joinable table with information if table has at least one row.
+        // This is needed because information_schema.table_rows is not reliable enough for innodb engine.
+        // see https://dev.mysql.com/doc/mysql-infoschema-excerpt/5.7/en/information-schema-tables-table.html TABLE_ROWS
+        $fromTableUnionSubSelectQuery = [];
+        foreach($tableNames as $tableName) {
+            $fromTableUnionSubSelectQuery[] = sprintf(
```
oliverklee, line 669: I propose we do this only for tables that don't have an auto increment (to improve performance).

lolli42: To my measurements, this does not make much of a difference. Both the version before and with this patch seem to be equally fast. Looks as if it's fine to avoid additional logic to separate this, here. What do you think?

oliverklee: If the performance is the same, then I'm fine with it. :-)

lolli42 (approved):

Classes/Core/Testbase.php
```diff
@@ -657,19 +653,47 @@ private function truncateAllTablesForMysql(): void
...
+        $fromTableUnionSubSelectQuery = implode(' UNION ', $fromTableUnionSubSelectQuery);
+        $query = sprintf("
+            SELECT
+                table_real_rowcounts.*,
+                information_schema.tables.AUTO_INCREMENT AS auto_increment
+            FROM (%s) AS table_real_rowcounts
+            INNER JOIN information_schema.tables ON (
+                information_schema.tables.TABLE_SCHEMA = %s
+                AND information_schema.tables.TABLE_NAME = table_real_rowcounts.table_name
+            )",
```
oliverklee, line 684: Do we need to use double quotes here?

Classes/Core/Testbase.php
```diff
@@ -657,19 +653,47 @@ private function truncateAllTablesForMysql(): void
...
-            $hasAtLeastOneRow = ((int)$tableData['Rows']) > 0;
+        foreach($result as $tableData) {
+            $hasChangedAutoIncrement = ((int)$tableData['auto_increment']) > 1;
+            $hasAtLeastOneRow = (bool)$tableData['has_rows'];
             $isChanged = $hasChangedAutoIncrement || $hasAtLeastOneRow;
             if ($isChanged) {
-                $tableName = $tableData['Name'];
+                $tableName = $tableData['table_name'];
                 $connection->truncate($tableName);
-                self::resetTableSequences($connection, $tableName);
```
oliverklee, line 672: Is removing this line intentional?

PR #265, opened 2021-08-28
Avoid truncate in database snapshotter

lolli42:


sbuerk (approved):
Looks good for me. Sure, we have to do more, but as next step approved.

PR #266, opened 2021-08-29
[TASK] Efficient sqlite db handling in functional tests

lolli42:
sqlite databases are no longer created in
typo3temp/var/tests/functional-"hash"/test.sqlite,
but instead in a parallel directory
typo3temp/var/tests/functional-sqlite-dbs/test_"hash".sqlite.

This allows defining this directory as tmpfs (eg. in runTests.sh),
which significantly improves functional test performance.

Additionally, when the first test created the database schema,
the db file is copied as test_"hash".empty.sqlite. Instead of
truncating the database in consecutive tests of the test case,
the "empty" file is simply copied around to avoid truncate queries.

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -283,13 +283,17 @@ abstract class FunctionalTestCase extends BaseTestCase
             self::$currestTestCaseClass = $currentTestCaseClass;
         }
 
+        // sqlite db path preparation
+        $dbPathSqlite = dirname($this->instancePath) . '/functional-sqlite-dbs/test_' . $this->identifier . '.sqlite';
+        $dbPathSqliteEmpty = dirname($this->instancePath) . '/functional-sqlite-dbs/test_' . $this->identifier . '.empty.sqlite';
+
         if (!$isFirstTest) {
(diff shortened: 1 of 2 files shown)


Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -283,13 +283,17 @@ protected function setUp(): void
...
             // Reusing an existing instance. This typically happens for the second, third, ... test
             // in a test case, so environment is set up only once per test case.
             GeneralUtility::purgeInstances();
             $this->container = $testbase->setUpBasicTypo3Bootstrap($this->instancePath);
             if ($this->initializeDatabase) {
-                $testbase->initializeTestDatabaseAndTruncateTables();
+                $testbase->initializeTestDatabaseAndTruncateTables($dbPathSqlite, $dbPathSqliteEmpty);
             }
             $testbase->loadExtensionTables();
         } else {
```
sbuerk, line 299: Should we not remove old sqlite database path on first instance creation, like it is done for instance files, since it is now seperated and coverd from the $this->instance_path anymore ? (Even if it is eventually in a ram disk ?)

sbuerk (approved):
Looks good to me.

PR #268, opened 2021-08-29
[BUGFIX] Restore post-processing of tables after a truncate

oliverklee:
This commit restores a line that accidentally was removed in #264.

### DIFF
--- a/Classes/Core/Testbase.php
+++ b/Classes/Core/Testbase.php
@@ -701,6 +701,7 @@ class Testbase
             if ($isChanged) {
                 $tableName = $tableData['table_name'];
                 $connection->truncate($tableName);
+                self::resetTableSequences($connection, $tableName);
             }
         }
     }


Classes/Core/Testbase.php
```diff
@@ -701,6 +701,7 @@ private function truncateAllTablesForMysql(): void
             if ($isChanged) {
                 $tableName = $tableData['table_name'];
                 $connection->truncate($tableName);
+                self::resetTableSequences($connection, $tableName);
```
lolli42, line 704: We actually dropped that by intention: This resetTableSequences() call is only needed for postgres, it's obsolete with mysql.

lolli42 (changes requested):

lolli42:
I'd keep it as is. Fine with you, Oliver?

Classes/Core/Testbase.php
oliverklee: Ah, okay. Thanks for the explanation!

Should we make this PostgreSQL-only then (in a separate PR)?

oliverklee: Ah, it already is.

lolli42: Unsure. There's a switch for sqlite in there, too. It's just that it was clearly obsolete for mysql/maria.

PR #269, opened 2021-08-29
[FEATURE] Keep development-only files out of the Composer packages

oliverklee:
This will reduce download size and improve download speed when this
package is used.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+/.Build/ export-ignore
+/.editorconfig export-ignore
+/.gitattributes export-ignore
+/.github/ export-ignore
+/.gitignore export-ignore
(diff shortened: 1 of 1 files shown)


oliverklee:
@lolli42 Should we exclude `Build/` from packaging as well, or are some of the files there used by the Core?

lolli42:
> @lolli42 Should we exclude `Build/` from packaging as well, or are some of the files there used by the Core?

Build/ can be excluded, too.

lolli42 (approved):

lolli42:
Thx. Looks good. I think there isn't anything possible than merging this, waiting for the next tagged version and see if something pops up. But I think it should be fine.

PR #272, opened 2021-08-29
[BUGFIX] Create PackageArtifact if needed

helhum:
Instead of manually disabling Composer mode by
overriding the Environment, create a
PackageArtifact for the desired packages if required
instead of the PackageStates.php file.


lolli42:
This looks great! Thank you!
Tested with styleguide core master & enetcache & core v10/v11.

lolli42 (approved):

PR #276, opened 2021-09-09
Fix PHP 8.1 compatibility.

Crell:

Since PHP 8.1, it's no longer possible to pass `$GLOBALS` by reference.  cf: https://wiki.php.net/rfc/restrict_globals_usage

The functional tests, however, do that on every single subrequest, which causes... a few thousand failures.  Fortunately there's an easy workaround.

(I'm not certain if this fixes all PHP 8.1 issues, but it is certainly the lion's share of them that can be found so far.)

Fixes: #

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -974,10 +974,13 @@ abstract class FunctionalTestCase extends BaseTestCase
         // carry that information.
         $_SERVER['X_TYPO3_TESTING_FRAMEWORK']['context'] = $context;
         $_SERVER['X_TYPO3_TESTING_FRAMEWORK']['request'] = $request;
-        ArrayUtility::mergeRecursiveWithOverrule(
-            $GLOBALS,
-            $context->getGlobalSettings() ?? []
-        );
+        // The $GLOBALS array may not be passed by reference, but its elements may be.
+        $override = $context->getGlobalSettings() ?? [];
+        foreach ($GLOBALS as $k => $v) {
+            if (isset($override[$k])) {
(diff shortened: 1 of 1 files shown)


oliverklee (approved):

sbuerk (approved):
Currently not seeing if it could have a sideeffekt.
For now it reduces a majorty core functional errors for php8.1, and also reduce the memory usages which rised for 8.1, not to that level of <8.1.

```
sysext:core   987 MB to 609 MB   on 8.1  | no change for 8.0
complete      4.68 GB => 1.54 GB on 8.1  | no change for 8.0
```

oliverklee (approved):

sbuerk (changes requested):
Sry Larry, may you to your PR a second replacement / fix ? Generally the same thing, but in another place.

Classes/Core/Functional/Framework/Frontend/RequestBootstrap
Line: 173 - 176

```php
ArrayUtility::mergeRecursiveWithOverrule(
    $GLOBALS,
    $this->context->getGlobalSettings() ?? []
);
```

with

```php
$override = $this->context->getGlobalSettings() ?? [];
foreach ($GLOBALS as $k => $v) {
    if (isset($override[$k])) {
        ArrayUtility::mergeRecursiveWithOverrule($GLOBALS[$k], $override[$k]);
    }
}
```

Both places together would reduce the functional error count for php81 to 17 (mariadb).

```
Tests: 4739, Assertions: 38070, Failures: 17, Skipped: 22, Incomplete: 7.
```

lolli42:
@sbuerk Let's take care of the executeFrontendRequest() related thing with a dedicated patch.
I'll merge the initial patch from Larry as is.

Issue #277, opened 2021-09-10
executeFrontendRequest() not usable in php8.1 as $GLOBALS are passed as reference


sbuerk:
Since php 8.1 passing $GLOBALS by reference is not supported anymore and leads to errors.

Core testing must use `executeFrontendSubRequest()` for v11 and onwards, and if not must be fixed
through core patch.

As there are existing extensions which uses the testing-framework for extension testing, it could be
that for moving on to v11, but not refactored tests yet, deprecated `executeFrontendRequest()` should
also work on php8.1 for now, until final removing from testing-framework.

PR #278, opened 2021-09-10
[BUGFIX] Avoid passing $GLOBALS by reference for executeFrontendRequest

sbuerk:
Passing $GLOBALS by reference is not supported since PHP 8.1.

This was fixed for v11 subRequests through #276, but exists for
executeFrontendRequest() also.

So far core master should use new subRequests, this should not be an issue
for core itself. Tests for v11+ must be fixed through as core patch moving
them to new subRequest testing.

Besides this extensions could and are using the testing-framework also for
testing it could be that moving forward to v11, but tests are still on
executeFrontedRequest() calls and not changed to executeFrontendSubRequests(),
and testing on php8.1. Thus this is fixed also, even if executeFrontend() will
be removed in the future from the testing-framework.

Fixes: #277


oliverklee (approved):

Crell (approved):
Looks good to me.  Is there anywhere else that $GLOBALS is passed directly that we should tackle while here?

PR #297, opened 2021-11-02
Fix typo

jdreesen:
I think it should be the same name as the one used in `post-autoload-dump`.


lolli42:
Hey. Thanks for your patch. The example is indeed broken. We'll no longer fix that in v8 and v9 branch, I'll cherry-pick your patch to master instead.

Issue #298, opened 2021-11-04
PHPStan: $coreExtensionsToLoad type has no value type specified in iterable type array.


DanielSiepmann:
### What are you trying to achieve?

I wanna have custom PHP code within my tests that does not result in errors when running phpstan on max level.

### What do you get instead?

```
 ------ -----------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   web/typo3conf/ext/extension/Tests/Functional/Page/GoogleTagManagerTest.php
 ------ -----------------------------------------------------------------------------------------------------------------------------------------------------------
  10     Property Vendor\Extension\Tests\Functional\Page\GoogleTagManagerTest::$coreExtensionsToLoad type has no value type specified in iterable type array.
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type
 ------ -----------------------------------------------------------------------------------------------------------------------------------------------------------
```

### How to reproduce the issue?

Add Property to your test class:
```php
<?php

namespace Vendor\Extension\Tests\Functional\Page;

use TYPO3\TestingFramework\Core\Functional\FunctionalTestCase;

class GoogleTagManagerTest extends FunctionalTestCase
{
    protected $coreExtensionsToLoad = [
        'fluid_styled_content',
        'filemetadata',
        'seo',
    ];
}
```

Execute PHPStan:
```sh
./vendor/bin/phpstan
```

With configuration:
```
parameters:
  level: max
  reportUnmatchedIgnoredErrors: true
  paths:
    - web/typo3conf/ext/
```

### Additional information you would like to provide?

Used Versions:
```
jangregor/phpstan-prophecy                      0.8.1                                   Provides a phpstan/phpstan extension for phpspec/prophecy
phpspec/prophecy-phpunit                        v2.0.1                                  Integrating the Prophecy mocking library in PHPUnit test cases
phpstan/extension-installer                     1.1.0                                   Composer plugin for automatic installation of PHPStan extensions
phpstan/phpstan                                 0.12.88                                 PHPStan - PHP Static Analysis Tool
phpunit/php-code-coverage                       9.2.6                                   Library that provides collection, processing, and rendering functionality for PHP code coverage infor...
phpunit/php-file-iterator                       3.0.5                                   FilterIterator implementation that filters files based on a list of suffixes.
phpunit/php-invoker                             3.1.1                                   Invoke callables with a timeout
phpunit/php-text-template                       2.0.4                                   Simple template engine.
[...]

PR #299, opened 2021-11-04
[BUGFIX] Properly annotate variable type of coreExtensionsToLoad

DanielSiepmann:

Fix PHPDoc annotation of property type.

1. Reproduce accordingly to issue https://github.com/TYPO3/testing-framework/issues/298
2. Clear cache of PHPStan, e.g. in `/tmp/phpstan`
3. Apply change and try to reproduce again, no errors should be shown this time.

### Results

I don't have that setup.

Fixes: #298

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -115,7 +115,7 @@ abstract class FunctionalTestCase extends BaseTestCase
      * A default list of core extensions is always loaded.
      *
      * @see FunctionalTestCaseUtility $defaultActivatedCoreExtensions
-     * @var array
+     * @var array<int, string>
      */
     protected $coreExtensionsToLoad = [];
 


Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -115,7 +115,7 @@ abstract class FunctionalTestCase extends BaseTestCase
      * A default list of core extensions is always loaded.
      *
      * @see FunctionalTestCaseUtility $defaultActivatedCoreExtensions
-     * @var array
+     * @var string[]
```
oliverklee, line 118: We can be even more specific:

```suggestion
     * @var array<int, string>
```

gilbertsoft: I've always thought this is exactly the same but you are right, the first one means `array<string|int, string>` see https://github.com/phpDocumentor/TypeResolver/blob/1.x/src/Types/AbstractList.php#L44.

oliverklee: … or `array<array-key, string>` (which semantically is the same, but communicates another intent).

lolli42: care to update to Olivers first remark?

PR #300, opened 2021-11-07
[TASK] Use non composer mode for extensions

gilbertsoft:
With the change https://review.typo3.org/c/Packages/TYPO3.CMS/+/72071
functional and acceptance tests are not longer working. Therefor the
mode is changed to non composer for these tests.


Classes/Core/SystemEnvironmentBuilder.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+/*
```
oliverklee, line 3: I propose we make new files strict to begin with.

gilbertsoft: Oops

Classes/Core/Testbase.php
```diff
@@ -164,29 +166,121 @@ public function removeOldInstanceIfExists($instancePath): void
...
+            }
+            if (!symlink($realTarget, $link)) {
+                throw new \RuntimeException(
+                    sprintf('Creating link failed (%s -> %s).', $link, $target),
                     1376657199
                 );
             }
         }
+
+        self::setUpEntryPoints($instancePath);
```
helhum, line 214: must be $this->

Classes/Core/Testbase.php
```diff
@@ -529,7 +623,12 @@ public function setUpPackageStates(
      */
     private function isPackageArtifactNeeded(): bool
     {
-        return defined('TYPO3_COMPOSER_MODE') && class_exists(PackageCacheEntry::class);
+        return false;
+
+        // For the time being we do override composer mode and this code is not
+        // needed anymore but kept here for later usage once real composer tests
+        // get included.
+        //return defined('TYPO3_COMPOSER_MODE') && class_exists(PackageCacheEntry::class);
```
helhum, line 631: We have version control. If we don't need code any more, we can safely remove it 😅

Classes/Core/Testbase.php
```diff
@@ -164,29 +166,121 @@ public function removeOldInstanceIfExists($instancePath): void
...
+
+        $this->defineOriginalRootPath();
+
+        /** @var array<string, string> $linksToSet */
         $linksToSet = [
-            '../../../../' => $instancePath . '/typo3_src',
-            'typo3_src/typo3' => $instancePath . '/typo3',
-            'typo3_src/index.php' => $instancePath . '/index.php',
+            ORIGINAL_ROOT => $instancePath . '/typo3_src',
+            'typo3_src/typo3/sysext/' => $instancePath . '/typo3/sysext',
```
helhum, line 192: Nothing we can tackle with this change, but this code will have a hard time when typo3 extensions are installed in vendor ...

Classes/Core/Testbase.php
```diff
@@ -164,29 +166,121 @@ public function removeOldInstanceIfExists($instancePath): void
...
-     * @return void
+     * @throws \InvalidArgumentException
+     * @throws \RuntimeException
      */
-    public function setUpInstanceCoreLinks($instancePath): void
+    public function setUpInstanceCoreLinks(string $instancePath): void
     {
+        $filesystem = new Filesystem();
+
+        if (!$filesystem->isAbsolutePath($instancePath)) {
```
helhum, line 180: is there a reason you introduce this check here now?

Classes/Core/Testbase.php
```diff
@@ -164,29 +166,121 @@ public function removeOldInstanceIfExists($instancePath): void
...
+     *
+     * @param string $instancePath Absolute path to test instance
+     * @throws \InvalidArgumentException
+     * @throws \UnexpectedValueException
+     */
+    public function setUpEntryPoints(string $instancePath): void
+    {
+        $filesystem = new Filesystem();
+
+        if (!$filesystem->isAbsolutePath($instancePath)) {
```
helhum, line 230: I don't think we need to double check this though

Classes/Core/Testbase.php
```diff
@@ -164,29 +166,121 @@ public function removeOldInstanceIfExists($instancePath): void
...
+        }
+
+        $entryPointsToSet = [
+            InstalledVersions::getInstallPath('typo3/cms-backend') . '/Resources/Private/Php/backend.php' => $instancePath . '/typo3/index.php',
+            InstalledVersions::getInstallPath('typo3/cms-frontend') . '/Resources/Private/Php/frontend.php' => $instancePath . '/index.php',
+            InstalledVersions::getInstallPath('typo3/cms-install') . '/Resources/Private/Php/install.php' => $instancePath . '/typo3/install.php',
+        ];
+
+        ;
+        if (($installersPath = InstalledVersions::getInstallPath('typo3/cms-composer-installers')) === null) {
```
helhum, line 244: we should either check for all path whether they are null or none of them imho

Classes/Core/Testbase.php
gilbertsoft: Yes, if it's not absolute we can get in trouble with symlinks created in the wrong folders.

gilbertsoft:
@helhum Found two other places where I have replaced the SysEnvBuilder.

bmack (approved):

helhum:
Thanks for the merge!

PR #313, opened 2021-12-01
[BUGFIX] Preserve locale in FrameworkState

ohader:

`locale` state shall be preserved in `FrameworkState` to avoid side-effects of simulating a frontend request in the very same PHP process.

Without this change running `\TYPO3\CMS\Frontend\Tests\Functional\SiteHandling\SiteRequestTest` fails - if locale `zh_CN.UTF-8` is not present in the corresponding test infrastructure.

---

### References

* https://bugs.php.net/bug.php?id=52923

### DIFF
--- a/Classes/Core/Functional/Framework/FrameworkState.php
+++ b/Classes/Core/Functional/Framework/FrameworkState.php
@@ -40,6 +40,7 @@ class FrameworkState
     public static function push()
     {
         $state = [];
+        $state['locale'] = setlocale(LC_ALL, 0);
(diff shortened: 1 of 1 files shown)


lolli42 (changes requested):
unsure on that. see my comment on https://review.typo3.org/c/Packages/TYPO3.CMS/+/72441

ohader:
I don't want to enforce this, since the side effects introduced by frontend sub requests are do not affect all operating systems..

PR #317, opened 2021-12-04
[BUGFIX] Avoid deprecated doctrine/dbal exception and method

sbuerk:

As preparation to raise doctrine/dbal to ^3.2 in core, we need to ensure
that removed things are replaced with corresponding methods. The replacements
are already available in 2.13.x, thus we can replace them beforehand, to stay
compatible with current used version, but have no blocker for the raise.

This pull-requests contains two commits, each commit doing only one thing.

1. Replace `\Doctrine\DBAL\DBALException` with `\Doctrine\DBAL\Exception`
2. Replace `exec()` with `executeStatement()`

> Both commits are marked to be backported to 7 to make future backports
> eventually a little bit easier.

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -585,7 +585,7 @@ abstract class FunctionalTestCase extends BaseTestCase
                     $sqlServerIdentityDisabled = false;
                     if ($platform instanceof SQLServerPlatform) {
                         try {
-                            $connection->exec('SET IDENTITY_INSERT ' . $tableName . ' ON');
+                            $connection->executeStatement('SET IDENTITY_INSERT ' . $tableName . ' ON');
                             $sqlServerIdentityDisabled = true;
                         } catch (DBALException $e) {
                             // Some tables like sys_refindex don't have an auto-increment uid field and thus no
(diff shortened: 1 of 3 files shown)


Classes/Core/Testbase.php
```diff
@@ -14,7 +14,6 @@
  * The TYPO3 project - inspiring people to share!
  */
 
-use Doctrine\DBAL\DBALException;
```
oliverklee, line 17: If we used this …:

```php
Doctrine\DBAL\Exception as DBALException;
```

… in all affected classes, we can leave the usages as-is, and still benefit from the short class names. What do you think?

sbuerk: Done.

Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -1140,7 +1140,7 @@ private function retrieveFrontendSubRequestResult(
      * + false: always deny, e.g. when importing data is finished
      *
      * @param bool|null $allowIdentityInsert
-     * @throws DBALException
+     * @throws \Doctrine\DBAL\Exception
```
oliverklee, line 1143: Generally for `@throws` annotations, it is not required to use a FQCN. So we can use `@throws DBALException` in all places.

sbuerk: Done.

sbuerk:
72508: [WIP][TASK] Test testing-framework fork patch
https://review.typo3.org/c/Packages/TYPO3.CMS/+/72508 (Patchset 1)

- `main|doctrine/dbal:2.13.x|pre-merge` https://git.typo3.org/typo3/CI/cms/-/pipelines/20566   green
- `main|doctrine/dbal:2.13.x|nightly` https://git.typo3.org/typo3/CI/cms/-/pipelines/20570   locked green, min/max red (expected due tf fork)

72372: [!!!][WIP][FEATURE] Require Doctrine/DBAL 3.2
https://review.typo3.org/c/Packages/TYPO3.CMS/+/72372 (Patchset 26)
- `main|doctrine/dbal:^3.2|pre-merge` https://git.typo3.org/typo3/CI/cms/-/pipelines/20565   green
- `main|doctrine/dbal:^3.2|nightly` https://git.typo3.org/typo3/CI/cms/-/pipelines/20567   locked green, min/max red (expected due tf fork)

PR #327, opened 2022-02-09
[BUGFIX] Still populate super globals $_GET, $_POST, $_REQUEST

ohader:
With commit d62c872d048015a3faee0b7b909302b8d3e54930 the frontend
sub request handling has been adjusted - basically only a PSR-7
request object is being used to transport request information.

However, the TYPO3 core (at that time, and currently) still relies
ob super globals like $_GET, $_POST and $_REQUEST.

Releases: main, 7, 6

### DIFF
--- a/Classes/Core/Functional/Framework/FrameworkState.php
+++ b/Classes/Core/Functional/Framework/FrameworkState.php
@@ -41,6 +41,9 @@ class FrameworkState
     {
         $state = [];
         $state['globals-server'] = $GLOBALS['_SERVER'];
+        $state['globals-get'] = $_GET;
+        $state['globals-post'] = $_POST;
+        $state['globals-request'] = $_REQUEST;
         $state['globals-beUser'] = $GLOBALS['BE_USER'] ?? null;
         // Might be possible to drop this ...
         $state['globals-typo3-conf-vars'] = $GLOBALS['TYPO3_CONF_VARS'] ?: null;
@@ -106,6 +109,10 @@ class FrameworkState
         $state = array_pop(self::$state);
 
         $GLOBALS['_SERVER'] = $state['globals-server'];
+        $_GET = $state['globals-get'];
+        $_POST = $state['globals-post'];
+        $_REQUEST = $state['globals-request'];
+
         if ($state['globals-beUser'] !== null) {
             $GLOBALS['BE_USER'] = $state['globals-beUser'];
         }
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -1001,6 +1001,18 @@ abstract class FunctionalTestCase extends BaseTestCase
         $requestUrlParts = parse_url((string)$request->getUri());
         $_SERVER['HTTP_HOST'] = $_SERVER['SERVER_NAME'] = $requestUrlParts['host'] ?? 'localhost';
 
+        if (isset($requestUrlParts['query'])) {
+            parse_str($requestUrlParts['query'], $_GET);
+            parse_str($requestUrlParts['query'], $_REQUEST);
+        }
+
+        if ($request->hasHeader('Content-Type')
+            // no further limitation to HTTP method, due to https://www.php.net/manual/en/reserved.variables.post.php
+            && in_array('application/x-www-form-urlencoded', $request->getHeader('Content-Type'))
+        ) {
+            parse_str((string) $this->request->getBody(), $_POST);
+        }
+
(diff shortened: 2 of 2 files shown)


ohader:
⚠️ Still WIP - let's check the impact...

ohader:
Let's see... → https://review.typo3.org/c/Packages/TYPO3.CMS/+/73376

lolli42:
what do you need this for?
in general i'd be fine with adding this hack to v6 TF (v11 core), but would love to avoid this in v12 - v12 core does not have *that* many _get / _post / _request dependencies anymore, so maybe we could see if we could get rid of the one that is a problem for you and triggers this patch for v12 here?
we've been working so hard on getting rid of hacks like this and the current functionals all work without setting this, it feels like a step back if we need to add this now. that's why im asking if we maybe could adapt core a bit to be a bit more clean ;)

ohader:
I totally understand the aim of using PSR-7 and proper request and response objects instead of accessing `$_GET` or `$_POST`. However, we are not there, **YET**.

The test-run in https://review.typo3.org/c/Packages/TYPO3.CMS/+/73376 uses this very pull-request an reveals different behavior in the frontend routing area - for instance in https://git.typo3.org/typo3/CI/cms/-/jobs/1034982

```
-        0 => 'https://blog.local/authors'
+        0 => '[https://blog.local/authors?type=0'](https://blog.local/authors?type=0%27)
```

The reason is, that we're still accessing `$_GET` in central places, like the following (there are some more)

https://github.com/TYPO3/typo3/blob/c3fa86f900ea13fca632266053aa6f9fa9bf115d/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L4109
https://github.com/TYPO3/typo3/blob/c3fa86f900ea13fca632266053aa6f9fa9bf115d/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L4928

Thus, the current test cases seem not to test the actual behavior, since `$_GET` and `$_POST` are not initialized.
This is kind of a blind spot in our tests - we think everything would be fine, but in reality it's probably not when using manual testing.

Once everything is settled for PSR-7 in the TYPO3 core, it's of course fine to drop these initializations. So, that's why I'd opt for reintroducing this PR for TYPO3 v11 and (temporarily) for TYPO3 v12.

lolli42:
Hmm. The two places you linked: COR usually gets the $request set since v11. So those could be easily adapted to $this->request?->getQueryParams() ?? [] in v12 I think. I'd thus prefer a core patch to see where it leads.

ohader:
My (personal) preferred approach would be:
* re-introduce workarounds for `$_GET` & `$_POST` - for TYPO3 v11 and TYPO3 v12
* analyze & adjust test cases that are failing after that - I guess(!) they reveal valid flaws in the TYPO3 core
* migrate to request objects in TYPO3 v12
* remove the re-introduced workarounds for TYPO3 v12 in testing framework

Observations in current TYPO3 core `main` branch (method calls via PhpStorm, skipped `Tests/` folder):

* `\TYPO3\CMS\Core\Utility\GeneralUtility::_GP` used 31 times
* `\TYPO3\CMS\Core\Utility\GeneralUtility::_GPmerged` used once
* `\TYPO3\CMS\Core\Utility\GeneralUtility::_GET` used 8 times
* `\TYPO3\CMS\Core\Utility\GeneralUtility::_POST` used 7 times
* total invocation of either `$_GET` or `$_POST` sums up to *47 times*

Referenced patch https://review.typo3.org/c/Packages/TYPO3.CMS/+/73372 just addresses two of them.

lolli42:
As far as i can see, both https://review.typo3.org/c/Packages/TYPO3.CMS/+/73372 and https://review.typo3.org/c/Packages/TYPO3.CMS/+/73376 find the exact same issues. So we need to fix these with https://review.typo3.org/c/Packages/TYPO3.CMS/+/73372 anyway. And after that, we have the situation that there is no difference between "This PR applied" and "This PR not applied" in core with our current coverage. So even if we merge it, we could revert that patch immediately after https://review.typo3.org/c/Packages/TYPO3.CMS/+/73372 has been merged to core again. I then suggest to just skip the PR.
Other than that, I'm totally with you: We should work on removing the last calls of that from the framework with fire. imho, that's a reachable goal in v12.

ohader:
Created PR for `6.x` branch to be used in TYPO3 v11 → https://github.com/TYPO3/testing-framework/pull/328
(making this very PR as WIP for the time being)

lolli42:
I hope it's ok to close here for now: Let's wait in v12 core to see what happens and ressurect this patch if really needed.

PR #334, opened 2022-02-25
[FEATURE] Provide access to private services in functional tests

lolli42:
Functional tests frequently need to instantiate $subject
or a dependency. This can be cumbersome if the service
in question is not public, since $this->getContainer()->get()
can currently only instantiate services marked as public.

The patch adds compiler passes as always-registered core
extension to DI in functional tests that allows to get()
private containers, too.

Also, FunctionalTestCase implements ContainerInterface,
this might be handy for some nifty cases where $this
can be hand over as container to a service to allow
custom callbacks if needed - Not a frequent option,
but useful in edge cases, use with care.

Based on:
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/TestServiceContainerWeakRefPass.php
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/TestServiceContainerRealRefPass.php
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Bundle/FrameworkBundle/Test/TestContainer.php

Related change for TYPO3 core:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/73646

Releases: main, v7, v6

### Results

Fixes: #

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -83,7 +83,7 @@ use TYPO3\TestingFramework\Core\Testbase;
  *     --bootstrap components/testing_framework/core/Build/FunctionalTestsBootstrap.php \
  *     typo3/sysext/core/Tests/Functional/DataHandling/DataHandlerTest.php
  */
-abstract class FunctionalTestCase extends BaseTestCase
+abstract class FunctionalTestCase extends BaseTestCase implements ContainerInterface
 {
     /**
      * An unique identifier for this test case. Location of the test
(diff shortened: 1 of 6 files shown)


Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -456,7 +457,12 @@ protected function getConnectionPool()
     }
 
     /**
+<<<<<<< HEAD
      * @return ContainerInterface
+=======
+     * @todo: Consider setting this private, too. So consumers have to
+     *        use get() / has().
+>>>>>>> d5d65dd... [FEATURE] Provide access to private services in functional tests
```
lolli42, line 465: oops.

lolli42:
will re-push.

PR #357, opened 2022-04-26
[TASK] Make it clear that processingfolder is not typo3temp

simonschaufi:
The previously set processingfolder starting with typo3temp could be confused with the actual typo3temp folder of TYPO3 which is not the same folder in reality as the folder in this case is a subdirectory of the file storage (fileadmin) in this case.

See https://review.typo3.org/c/Packages/TYPO3.CMS/+/73932 which brought up this confusion.

### DIFF
--- a/Resources/Core/Functional/Fixtures/sys_file_storage.xml
+++ b/Resources/Core/Functional/Fixtures/sys_file_storage.xml
@@ -4,7 +4,7 @@
 		<uid>1</uid>
 		<pid>0</pid>
 		<name>fileadmin/ (auto-created)</name>
-		<processingfolder>typo3temp/assets/_processed_/</processingfolder>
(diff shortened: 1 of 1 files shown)


simonschaufi:
No adjustments needed when merging this into TYPO3 core.

sbuerk:
Granted, that lead to some confusion (on my side) during a core patch review. However, I'm not sure what was the intension to youse this path as processing folder. Maybe it was another. Will look into it on the weekend. 

This is still "somehow" confussing, even if not typo3temp is used. And not checked if eventually some tests may rely on this somehow. So no approval for now.

simonschaufi:
I have checked it and the core doesn't depend on this path.

sbuerk (approved):

PR #360, opened 2022-05-03
[!!!][TASK] Remove MSSQL supportive code

sbuerk:
Core requirements have been changed and publicly announced with:
https://typo3.org/article/new-system-requirements-for-upcoming-typo3-v12

This patch removes supportive code for MSSQL due dropped support
for this dbmns.

* remove mssql special handling code for ident columns in several places
* remove wrapped connection which was needed to intercept for special
  mssql ident column handling
* remove mssql special handling for auto increment value handling used
  in test related database handling code
* check for mssql drivers and throw a proper exception to early inform
  about dropped support instead of let users run in mysterious errors

Releases: main

Verified against core pre-merge + nightly in: https://review.typo3.org/c/Packages/TYPO3.CMS/+/74497


sbuerk:
We should think about/discuss backporting only the new exception/driver check in a dedicated patch on branch 7 with a core version check (v11 okay, v12 exception).

Classes/Core/Testbase.php
```diff
@@ -427,6 +426,17 @@ public function getOriginalDatabaseSettingsFromEnvironmentOrLocalConfiguration(a
                 1397406356
             );
         }
+
+        // Instead of letting code run and retrieving mysterious result and errors, check for dropped sql driver
+        // support and throw a corresponding exception with a proper message.
+        if (in_array($originalConfigurationArray['DB']['Connections']['Default']['driver'], ['sqlsrv', 'pdo_sqlsrv'], true)) {
```
oliverklee, line 432: Do we need to have an `??` operator here in order to avoid invalid array accesses for broken setups?

sbuerk: Will add one. In the end, this would only be the case when typo3conf/LocalConfiguraiton exists and is used. ("parent instance"). Whould should have a db setup if install has run.

Will add the ?? operator later.

lolli42 (approved):

PR #362, opened 2022-05-05
[BUGFIX] Add the PDO PHP extension as a requirement

oliverklee:
As we are using the `PDO` class, we need to have it as an explicit
requirement in order to get a proper warning when the extension
is missing (instead of random breakage).

Releases: main, 7, 6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
   },
   "require": {
     "php": "^8.1",
+    "ext-pdo": "*",
     "phpunit/phpunit": "^9.5.10",
     "psr/container": "^1.1.0 || ^2.0.0",
     "mikey179/vfsstream": "~1.6.10",


sbuerk:
Hi Oliver, 

I would avoid this patch. Granted you are right, in the currently shape. Not sure, but I think had that topic lately with someone. 

First, we should replace the type constants from the core Connection facade class from `PDO::XXX` to either doctrine/dbal `ParameterType::XXX` alias constants or direct integer (prefered the ParameterType one).

And afterwards the usage in code for parameters we should replace them through core, testing-framework and styleguide to use the `Connection::xxx` instead of `PDO::xxx`.

And avoid hard requiring a php extension which should be "optional".

oliverklee:
@sbuerk Thanks!

What do you think about this approach:
1. add the PDO extension as a requirement as long as we are using it (which means that is will be a transitive dev requirement in the core as the testing framework is used as a dev requirement there)
2. have another PR after that that converts the usages of the `PDO::` constants to something else and removes the requirement again?

sbuerk:
I would say - should not be needed. core has for monorepo root and for "typo3/cms-core" already ext pdo has "hard requirement" - maybe not working due typo ?

https://github.com/TYPO3/typo3/blob/252173d3449c44121cd7dfa5af3ee981278d26c7/composer.json#L40
https://github.com/TYPO3/typo3/blob/d77c96608a9192a0ea6a762e73afc2b1acf84c09/composer.json#L40
https://github.com/TYPO3/typo3/blob/6107c4fdab38b87533915cccf5536056f79942bb/composer.json#L40

https://github.com/TYPO3/typo3/blob/252173d3449c44121cd7dfa5af3ee981278d26c7/typo3/sysext/core/composer.json#L23
https://github.com/TYPO3/typo3/blob/d77c96608a9192a0ea6a762e73afc2b1acf84c09/typo3/sysext/core/composer.json#L23
https://github.com/TYPO3/typo3/blob/6107c4fdab38b87533915cccf5536056f79942bb/typo3/sysext/core/composer.json#L23

TF requires typo3/cms-core, thus this requirement as all are requirements and not dev requirements.

Maybe this have been a long oversight (in core already). `pdo` is stated as server requirements, so we should:

1. first check if "ext-PDO" has not worked for check
2. fix it in core

no need for adjustment in TF.

Usage may be migrated anyway, but as PDO is hard requirement not urgent. Fixing it meanwhile in core - the better way.

sbuerk (approved):

lolli42:
I'd say this discussion is easy to solve: ext:core has this dependency, too. So it does not matter much if we add it or not, since missing php pdo extension would then fail with ext:core. I think we should simply add it here, too.

PR #363, opened 2022-05-05
[TASK] Add PHPUnit rules for PHPStan

oliverklee:
Also update PHPStan to the latest version.

Releases: main, 7, 6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,8 @@
   },
   "require-dev": {
     "typo3/coding-standards": "^0.5.0",
-    "phpstan/phpstan": "^1.4.9",
+    "phpstan/phpstan": "^1.6.7",
+    "phpstan/phpstan-phpunit": "^1.1.1",
     "typo3/cms-workspaces": "12.*.*@dev"
   }
 }


oliverklee:
This PR blocks another work-in-progress change of mine which makes the PHPDoc type annotations more specific.

lolli42:
I guess you're already aware the config is not auto-loaded. It should be added to phpstan.neon similar like the 'include' section in core-monorepo does: https://github.com/TYPO3/typo3/blob/main/Build/phpstan/phpstan.neon
I assume you'll do that with the next patch and would go ahead and merge this patch now.

lolli42:
I guess you're already aware the config is not auto-loaded. It should be added to phpstan.neon similar like the 'include' section in core-monorepo does: https://github.com/TYPO3/typo3/blob/main/Build/phpstan/phpstan.neon
I assume you'll do that with the next patch and would go ahead and merge this patch now.

sbuerk (approved):

oliverklee:
> I guess you're already aware the config is not auto-loaded. 

I actually forgot that. Thanks for the reminder! I'll provide a PR for loading it.

oliverklee:
Done: #366

PR #364, opened 2022-05-05
[TASK] Use more native types in the test base classes

oliverklee:


sbuerk:
in that form "main" only

oliverklee:
> in that form "main" only

Yes, that's was I was thinking as well. :-)

sbuerk:
> > in that form "main" only
> 
> Yes, that's was I was thinking as well. :-)

Yeah .. even no `7` backport in that way. union/intersection types not possible below php8.0/php8.1 thus not working with core v11/php7.4-php8.1 range. So maybe eventually partly.

oliverklee:
Then let's keep this main-only.

sbuerk (approved):

PR #365, opened 2022-05-06
[!!!][TASK] Use session identifier JWT when initializing sessions

ohader:
see https://review.typo3.org/c/Packages/TYPO3.CMS/+/69337

Frontend and Backend Cookies are not just the plain session identifier anymore, but the cookie values a hash-signed JWT containing the corresponding session identifier.


sbuerk:
just to note - as the core patch is "marked" for core v11, this have to be backported to 7 and 6. And in 6 it must be done conditional based, as TF 6.x is used for v10 and v11 testing.

lolli42 (approved):

sbuerk (approved):

PR #373, opened 2022-05-31
[TASK] Update to PHPStan 1.7.10

oliverklee:
Used commands:

```bash
composer req --dev phpstan/phpstan:^1.7.7
./Build/Scripts/runTests.sh -s phpstanGenerateBaseline
```

Releases: main, 7, 6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
   },
   "require-dev": {
     "typo3/coding-standards": "^0.5.0",
-    "phpstan/phpstan": "^1.7.0",
+    "phpstan/phpstan": "^1.7.10",
     "phpstan/phpstan-phpunit": "^1.1.1",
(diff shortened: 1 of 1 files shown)


oliverklee:
Marked as WIP as there are no baseline changes and we could wait for one of the next PHPStan releases.

lolli42:
hey. i'll abandon here and will just go with a 1.8.0 raise.

lolli42:
note @oliverklee please don't push PR's like these when they don't add significant benefit: Those take like around half an hour to manage with backports to get right and be sure about it. I'd strive to raise phpstan and friends occassionally when needed because it gives real benefits, only.

oliverklee:
@lolli42 Is the half hour the review time or the build time on CI? (I usually mark those PRs as draft so nobody will put time into review/merge/backports).

PR #374, opened 2022-06-06
[BUGFIX] Drop native return types for `getAccessibleMock` & friends

oliverklee:
The combination of "union types with generics in PHPDoc" and
"a subset of those types as native types, but without the generics"
tends to confuse PhpStorm's static type analysis (which we want to avoid).

So we stay with PHPDoc return type annotations in the affected places.

Releases: main


Classes/Core/BaseTestCase.php
```diff
@@ -30,6 +30,9 @@ abstract class BaseTestCase extends TestCase
     /**
      * Creates a mock object which allows for calling protected methods and access of protected properties.
      *
+     * Note: This method has no native return types as having native (incomplete) return types, PHPDoc return type
+     * annotations and generics in combination tends to confuse PhpStorm's static type analysis.
```
sbuerk, line 34: Mhh .. this sentens sounds kind weired. "has no native return typas as having native (inclomplete) return types..." and so go ...

Classes/Core/BaseTestCase.php
```diff
@@ -81,6 +84,9 @@ protected function getAccessibleMock(
      * of protected properties. Concrete methods to mock can be specified with
      * the last parameter
      *
+     * Note: This method has no native return types as having native (incomplete) return types, PHPDoc return type
```
sbuerk, line 87: same as above - sounds weired.

sbuerk (changes requested):
May you be up to rephrase the two (same) comments ? Otherwise change/PR looks good and works so far.

lolli42:
how to continue here? if monorepo-main is fine with this, i'd tend to be fine with it, too? maybe rephrase wording a bit?

oliverklee:
Yes, I‘ll update the wording (probably on Saturday, i.e., tomorrow).

Classes/Core/BaseTestCase.php
oliverklee: I have rephrased those two comments now. I hope that they now are less weird and less confusing. :-)

Issue #375, opened 2022-06-13
AbstractModalDialog not compatible with Bootstrap 5


saitho:
### What are you trying to achieve?

TYPO3 CMS 11 upgraded Bootstrap from v3 to v5. 
Since Bootstrap 4 active modals have the "show" class, instead of the "in" class.
AbstractModalDialog still uses the "in" class.

### What do you get instead?

Acceptance tests on TYPO3 CMS 11 using AbstractModalDialog fail.

> 1) DividerCest: Can create via list module
 Test  vendor/our-tests/tests/acceptance_backend/ContentElements/DividerCest.php:canCreateViaListModule
>                                                    
>  [Facebook\WebDriver\Exception\TimeoutException]   
>                                                    
>
>Scenario Steps:
>
> 28. $I->waitForElement(".modal.in") at vendor/typo3/testing-framework/Classes/Core/Acceptance/Helper/AbstractModalDialog.php:70
> 27. $I->switchToIFrame() at vendor/typo3/testing-framework/Classes/Core/Acceptance/Helper/AbstractModalDialog.php:69

### How to reproduce the issue?

We test that divider CE can be created via list module:
```php
public function canCreateViaListModule(BackendTester $I, ModalDialog $modalDialog): void
{
    $I->useExistingSession('editor');

    $I->click('#web_layout');
    $pageTree->openPath([
        'Testfälle'
    ]);
    $I->comment('I select the page "Content Elemente"');
    $I->click('g#identifier-0_7');
    $I->switchToFrame();
    $I->click('#web_list');

    $I->switchToContentFrame();
    $I->waitForElement('.module-docheader-bar [data-identifier="actions-add"]');
    $I->click('.module-docheader-bar [data-identifier="actions-add"]');
    $I->waitForText('New record');
    $I->click('Page Content');

    $I->waitForElementVisible('.typo3-TCEforms');
    $I->selectOption('select[name*="[CType]"]', 'div');
    $modalDialog->clickButtonInDialog('OK');
}
```

### Specify some data of the environment

- TYPO3 testing framework version: 6.15.2
- TYPO3 version: 11.5.8
- TYPO3 installation type: composer
- PHP version: 7.4 via DDEV
- Selenium server version: Docker Image `selenium/standalone-chrome:3`
- Browser driver (chromedriver/geckodriver...) version: 
- Browser used + version: chrome

PR #376, opened 2022-06-13
[BUGFIX] Use Bootstrap 5 modal class

saitho:

Since Bootstrap 4 active modals have the "show" class, instead of the "in" class.
This fix restores compatibility with Bootstrap 5 which is used since TYPO3 CMS 11.

Running a test that uses ModalDialog (see example in #375) with TYPO3 CMS 11.

### Results

The test should succeed.

Fixes: #375


lolli42:
core main is happy with this: https://review.typo3.org/c/Packages/TYPO3.CMS/+/75019

i'll merge to main and v7

Issue #377, opened 2022-06-26
Static Data in Database is not kept/recreated for second FunctionalTest


sHorst:
### What are you trying to achieve?

Run more than one test requiring Static data in one TestClass Class

### What do you get instead?

Error for missing data in static tables (in my case static_country)

### How to reproduce the issue?

```php
class PlatzTest extends FunctionalTestCase {
    
    [...]

    /**
     * @test
     */
    public function searchOverviewAction()
    {
        $response =  $this->executeFrontendSubRequest(
            (new InternalRequest())->withQueryParameters(
                [
                    'id' => self::PLATZ_CONTROLLER_PID,
                    self::PLUGIN_NAME . '[action]' => 'search',
                    self::PLUGIN_NAME . '[controller]' => 'Platz',
                ]
            )
        );

        $content = (string)$response->getBody();

        self::assertMatchesRegularExpression('#<a href=".*"> Bayern</a>#', $content);
        self::assertMatchesRegularExpression('#<a href=".*"> Nordrhein-Westfalen</a>#', $content);
    }

    /**
     * @test
     */
    public function searchBundeslandAction()
    {
        $respones = $this->executeFrontendSubRequest(
            (new InternalRequest())->withQueryParameters(
                [
                    'id' => self::PLATZ_CONTROLLER_PID,
                    self::PLUGIN_NAME . '[action]'  => 'search',
                    self::PLUGIN_NAME . '[controller]' => 'Platz',
                    self::PLUGIN_NAME . '[countryZone]' => '88',
                ]
            )
        );

        $content = (string)$respones->getBody();

        // Check if campsite 1 is listed
        self::assertMatchesRegularExpression('#Campsite 1564940472#', $content);
    }
}
```

Relevant code from the Tested Class:

```php
  /**
     * @return array
     */
    private function _getAllCountryZones(): array
    {
        /** @var \SJBR\StaticInfoTables\Domain\Model\Country $de */
        $de = $this->countryRepository->findByUid($this->settings['germanyUid']??54);
        $countryZones = $this->countryZoneRepository->findByCountry($de)->toArray();

        // sort by name
        usort($countryZones, function (CountryZone $a, CountryZone $b) {
            return strnatcasecmp($a->getLocalName(), $b->getLocalName());
        });
        return $countryZones;
    }
```

### Additional information you would like to provide?

[...]

PR #378, opened 2022-06-26
reimport static content after truncate

sHorst:

Reimport Static data into database after truncating All Tables

Run two TestCases in one class which use Static Data

### Results

Static Data is not missing for Second Test and therefore can be Tested

Fixes: #377

### DIFF
--- a/Classes/Core/Testbase.php
+++ b/Classes/Core/Testbase.php
@@ -671,6 +671,14 @@ class Testbase
         } else {
             $this->truncateAllTablesForOtherDatabases();
         }
+
+        // reimport static content
+        $schemaMigrationService = GeneralUtility::makeInstance(SchemaMigrator::class);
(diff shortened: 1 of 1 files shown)


lolli42:
Not 100% sure here, this may need further investigation:
How expensive is it to call the db analyzer for each test and could we maybe combine it with the 'snapshot' system that is done for sqlite?

sHorst:
as i understand, the class is not destroyed while testing, the analyzed include statements could be stored in a class variable as well. This has then no timeing impact, but only memory consumption issues.

But you are more into the details of the function, so i am open to suggestions. This was a "easy" way to fix the issue.

lolli42:
Hey.
The decision is to drop the auto-import of ext_tables_static+adt.sql files within createDatabaseStructure() altogether. See the issue for more information on this.
As such, this PR is not gonna make it as proposed and we'll close it now.

PR #430, opened 2022-11-30
[TASK] Ensure fixture field data for json fields can be inserted

sbuerk:
With https://review.typo3.org/c/Packages/TYPO3.CMS/+/76553
support for native json database field has been introduced 
into TYPO3. This came along with automatic transformation 
from PHP-Value to database compatible values, which basicly 
means that the values get json-encode'd.

Providing fixture database values for JSON fields are now 
impossible to set, as string values get encoded as string
which does not write the expected values to the database.

This change uses the already used database schema information
and proper decode the values after reading from the csv files.
That ensures that the data written to the database has the
expected format. Othewise testing would be inpossible for
JSON-Fields.

Releases: main, 7

### DIFF
--- a/Classes/Core/Functional/Framework/DataHandling/DataSet.php
+++ b/Classes/Core/Functional/Framework/DataHandling/DataSet.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 
 namespace TYPO3\TestingFramework\Core\Functional\Framework\DataHandling;
 
+use Doctrine\DBAL\Types\JsonType;
 use TYPO3\CMS\Core\Database\ConnectionPool;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\TestingFramework\Core\Testbase;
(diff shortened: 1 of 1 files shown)


sbuerk:
Backport to 7 needs a kindly other solution. Will take care later of the backport in a dedicated PR.

Classes/Core/Functional/Framework/DataHandling/DataSet.php
```diff
@@ -70,13 +71,23 @@ public static function import(string $path): void
         $dataSet = self::read($path, true);
         foreach ($dataSet->getTableNames() as $tableName) {
             $connection = GeneralUtility::makeInstance(ConnectionPool::class)->getConnectionByName(ConnectionPool::DEFAULT_CONNECTION_NAME);
+            $platform = $connection->getDatabasePlatform();
```
bmack, line 74: I think we can move the two lines out of the outer foreach()

sbuerk: Done

o-ba (approved):

Issue #444, opened 2023-03-20
AccessibleMock incompatible with phpunit 10


liayn:
### What are you trying to achieve?

`$this->getAccessibleMock(SomeClass::class)`

Should work.

### What do you get instead?

>Error: Call to undefined method PHPUnit\Framework\MockObject\MockBuilder::setMethods()
> /builds/.../.Build/vendor/typo3/testing-framework/Classes/Core/BaseTestCase.php:59

The method `MockBuilder::setMethods` is gone in phpunit v10.

### Specify some data of the environment

- TYPO3 testing framework version: 7.x-dev aca2f49
- TYPO3 version: 11.5.25
- TYPO3 installation type: composer
- PHP version: 8.2
- phpunit/phpunit (10.0.16)

### related links

https://github.com/sebastianbergmann/phpunit/pull/3687
https://medium.com/bumble-tech/working-with-partial-mocks-in-phpunit-644637e12f84
https://www.drupal.org/project/drupal/issues/3130606

lolli42:
In Fluid standalone, I switched these to setMethods() to onlyMethods() ... guess we should do the same here. The difference is, that onlyMethods() excepts when a methods to mock does not exist, while setMethods() does not. Various core tests that use "dummy" will need to be adapted.

liayn:
The other way around ;-) `setMethods` excepts non-existing methods, whereas the newer `onlyMethods` doesn't.
Those dummy-methods now need `addMethods` to be used.

I guess this is indeed a breaking change for the "accessible proxy" API of the testing framework, unless we do this detection ourselves, but I tend to push this responsibility to the extension author.

PR #446, opened 2023-03-27
[TASK] Avoid setMethods() on mocks

lolli42:
setMethods() has been removed with phpunit 10.

$this->getAccessibleMock() used this to only
partially mock a test subject. The patch switches
to onlyMethods() now.

There is one difference: onlyMethods() fails if
a methods that should be mocked does not exist,
while setMethods() did not fail on this.

So this is more strict now. It was frequently used in core tests mocking ['dummy'] to say "mock no method". Those have to be switched to use 'null' as argument to $this->getAccessibleMock() now. Also, when a class is refactored, and a method that has been previously mocked is removed with the refactoring, the mock has to be adapted to remove the method-mock request from the test as well now, otherwise the test will fail.

Releases: main, 7
Resolves: #444

### DIFF
--- a/Build/phpstan/phpstan-baseline.neon
+++ b/Build/phpstan/phpstan-baseline.neon
@@ -1,11 +1,6 @@
 parameters:
 	ignoreErrors:
 		-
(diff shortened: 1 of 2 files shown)


lolli42:
let's keep it pending for a while to see if this is the way to go when fixing core, or if this needs / should have furher quirks.

sbuerk (dismissed):

Issue #448, opened 2023-04-05
Remember to adapt .xml phpunit files when main goes phpunit 10 only


lolli42:

liayn:
Can we migrate right away?

Because for my extensions the CI is now failing since this is a warning now and causes exit-code 1.

```
$ .Build/bin/phpunit --configuration .Build/vendor/typo3/testing-framework/Resources/Core/Build/UnitTests.xml --colors=never Tests/Unit/
PHPUnit 10.0.19 by Sebastian Bergmann and contributors.
Runtime:       PHP 8.2.4
Configuration: .Build/vendor/typo3/testing-framework/Resources/Core/Build/UnitTests.xml
................................................................. 65 / 80 ( 81%)
...............                                                   80 / 80 (100%)
Time: 00:00.334, Memory: 18.00 MB
There was 1 PHPUnit warning:
1) Your XML configuration validates against a deprecated schema. Migrate your XML configuration using "--migrate-configuration"!
WARNINGS!
```

lolli42:
Please don't use that file directly, but copy to your consuming project / extension.

For 7 branch (core compat 11 & 12), we'll probably add a second boilerplate file for -10.xml as example, for main branch (core compat 12), we'll adapt the existing file as soon as phpunit 9 has been removed from composer.json.

PR #450, opened 2023-04-06
[TASK][#448] Add phpunit configuration for v10

liayn:


sbuerk:
Thanks, but I would postpone this currently for now.

First, files are templates and phpunit can convert it. So you could easily do it in your project/extension directly.

Second, current plan is to hard switch to phpunit10 and not supporting a dualmode. So adding this as intermediate may be more confusing in the end than.

Third, we want to add some more phpunit stuff to mitigate things phpunit has dropped (deprecation to exceptions and therefore failing tests), avoiding auto converting @deprecation annotations to trigger_error() calls on class mocks, despite if the method is called or not and so on.

PR #435, opened 2023-01-28
[TASK] Ensure "typo3/cms-composer-installers" v5 compatibility

sbuerk:
Since TYPO3 v12.0 the "typo3/cms-composer-installers" package
is a hard requirement in version 5. This comes along with some
changes:

* system-extensions are no longer installed in
  `<public-dir>/typo3/sysext/*`

* extensions are no longer installed in
  `<public-dir>/typo3conf/ext/*`

Extension and System Extensions are now installed into the
`vendor-dir` folder, which is the default behaviour of the
composer package-manager. Generally, this is really good
move.

"typo3/testing-framework" expected a specific setup structure,
to provide correct symlinks for system and extensions for
functional- and acceptance test instances. Which is not given
anymore.

As a intermediate solution, a composer plugin has been put
into the wild, providing the the expected structure with
symlinks. Which is a horrable solution, but worked for the
meanwhile.

This change now introduces a testing-framework internal
composer package manager to retrieve informations about
root packags and required packages and use the loaded
information for symlink handling.

Benefits of the implementation:

* core- and test extensions to load can now be specified by
  using the extension-key and/or the composer package name.
  `typo3conf/ext/extension-key` compatibilty is added and
  works without changes. May be deprecated at a later point.

* absolute paths to test-fixture extensions still works the
  same way

* testing-framework do not longer expects system-extensions
  or extensions in a specific place

* works with cms-composer-installers 3, 4RC1 and 5 the same
  way

* "sbuerk/typo3-cmscomposerinstallers-testingframework-bridge"
  is obsolete now, therefore added a "replace" for it to
  `composer.json`

This restores working state. With further dedicated change, this
could be extended by adding a composer.json file scanner to find
and register `test-fixture extensions` directly. Which further
forces that extensions should have always a composer.json.

Releases: main, 7
Resolves: #403

--------------------------------------------------------------------------

tf-main (#434)

* [77624: [WIP][TASK] Test TF typo3/cms-composer-installers 5 changes](https://review.typo3.org/c/Packages/TYPO3.CMS/+/77624)
* [typo3/styleguide (PR)](https://github.com/TYPO3/styleguide/pull/373)

tf-7 (#435)

* [sbuerk/typo3-ensure-admin (PR)](https://github.com/sbuerk/typo3-ensure-admin/pull/6)
* [lolli42/dbdoctor (PR)](https://github.com/lolli42/dbdoctor/pull/26)

--------------------------------------------------------------------------

[...]

PR #461, opened 2023-04-23
Adjust variables, use composer info, allow `EXT:` in paths

DavidBruchmann:
Introduce Composer/ComposerPackageInfo to provide package info and selection based on composer and related to TYPO3. Furthermore extensions can be linked with prefix `EXT` and the path is resolved.

adjust Testbase for TYPO3 v12 and the new class ComposerPackageInfo

This PR is partially in competition to #435 (and perhaps other PRs), but I see it as exchange of code to find the best way. So some of my ideas could be taken for #435 or some other kind of merge could be considered.

- resolve paths fore independent on file structure
- filter composer packages for TYPO3 system or 'local' extensions
- allow incoming configuration to use the prefix `EXT:` for extensions
- change how symlinks to core extensions are created:
    before the whole sysext folder was linked as one symlink, now the sysext folder is created and each extension symlinked individually.

- install https://github.com/DavidBruchmann/t3docs-screenshots/ when it's updated
- execute the basic test like described in the README (without need to load external documentation).

### Results

All suite installations are working properly.
Main installation and acceptance tests never work yet, due to other reasons).

### DIFF
--- a/Classes/Composer/ComposerPackageInfo.php
+++ b/Classes/Composer/ComposerPackageInfo.php
@@ -188,6 +188,6 @@ class ComposerPackageInfo extends \Composer\InstalledVersions
 
     public static function getAutoload()
     {
-        return !empty($_composer_autoload_path) ? $_composer_autoload_path : realpath(InstalledVersions::getInstallPath('typo3/cms-core') . '/../../autoload.php');
+        return realpath(InstalledVersions::getInstallPath('typo3/cms-core') . '/../../autoload.php');
     }
 }
--- a/Classes/Composer/ComposerPackageInfo.php
+++ b/Classes/Composer/ComposerPackageInfo.php
@@ -188,6 +188,6 @@ class ComposerPackageInfo extends \Composer\InstalledVersions
 
     public static function getAutoload()
     {
-        return @isset($_composer_autoload_path) ? $_composer_autoload_path : realpath(InstalledVersions::getInstallPath('typo3/cms-core') . '/../../autoload.php');
+        return !empty($_composer_autoload_path) ? $_composer_autoload_path : realpath(InstalledVersions::getInstallPath('typo3/cms-core') . '/../../autoload.php');
     }
 }
--- a/Classes/Composer/ComposerPackageInfo.php
+++ b/Classes/Composer/ComposerPackageInfo.php
@@ -16,7 +16,7 @@ namespace TYPO3\TestingFramework\Composer;
  * The TYPO3 project - inspiring people to share!
  */
 
-use \Composer\InstalledVersions;
+use Composer\InstalledVersions;
 
 /**
  * PROPERTIES of \Composer\InstalledVersions:
--- a/Classes/Composer/ComposerPackageInfo.php
+++ b/Classes/Composer/ComposerPackageInfo.php
@@ -184,6 +184,6 @@ class ComposerPackageInfo
 
     public static function getAutoload()
     {
-        return realpath(self::getInstallPath('typo3/cms-core') . '/../../autoload.php');
+        return realpath(InstalledVersions::getInstallPath('typo3/cms-core') . '/../../autoload.php');
     }
 }
--- a/Classes/Composer/ComposerPackageInfo.php
+++ b/Classes/Composer/ComposerPackageInfo.php
@@ -1,0 +1,191 @@
+<?php
+
+declare(strict_types=1);
+namespace TYPO3\TestingFramework\Composer;
+
+/*
+ * This file is part of the TYPO3 CMS project.
+ *
+ * It is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License, either version 2
+ * of the License, or any later version.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.txt file that was distributed with this source code.
+ *
+ * The TYPO3 project - inspiring people to share!
+ */
+
+/**
+ * PARENT PROPERTIES:
+ * private static $installed;
+ * private static $canGetVendors;
+ * private static $installedByVendor = array();
+ *
+ * PARENT METHODS:
+ * public static function getInstalledPackages()
+ * public static function getInstalledPackagesByType($type)
+ * public static function isInstalled($packageName, $includeDevRequirements = true)
+ * public static function satisfies(VersionParser $parser, $packageName, $constraint)
+ * public static function getVersionRanges($packageName)
+ * public static function getVersion($packageName)
+ * public static function getPrettyVersion($packageName)
+ * public static function getReference($packageName)
+ * public static function getInstallPath($packageName)
+ * public static function getRootPackage()
+ * public static function getRawData() @deprecated
+ * public static function getAllRawData()
+ * public static function reload($data)
+ * private static function getInstalled()
+ */
+
+/**
+ * TYPO3 related extending class of \Composer\InstalledVersions.
+ * @see https://github.com/composer/composer/blob/main/src/Composer/InstalledVersions.php
+ */
+class ComposerPackageInfo extends \Composer\InstalledVersions
+{
+    /**
+     * Get's all TYPO3 local extensions that are installed.
+     * This is ignoring system extensions.
+     */
+    public static function getLocalExtensions()
+    {
+        return self::getInstalledPackagesByType('typo3-cms-extension');
+    }
+
+    /**
+     * Get's all TYPO3 system extensions that are installed.
+     * This is ignoring local extensions.
+     */
+    public static function getSystemExtensions()
+    {
+        return self::getInstalledPackagesByType('typo3-cms-framework');
+    }
+
+    /**
+     * Get's all TYPO3 extensions that are installed.
+     * This is ignoring i.e. packages that serve as composer plugin or are independent libraries
+     */
+    public static function getAllExtensions()
+    {
+        $localExtensions = self::getLocalExtensions();
+        $systemExtensions = self::getSystemExtensions();
+        return array_merge($systemExtensions, $localExtensions);
+    }
+
+    /**
+     * Get's all TYPO3 local extensions that are installed.
+     * This is ignoring system extensions.
+     */
+    public static function getLocalExtensionsEnriched()
+    {
+        $localExtensions = self::getLocalExtensions();
+        $localExtensionsEnriched = self::getExtensionsEnriched($localExtensions);
+        return $localExtensionsEnriched;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


sbuerk:
Not sure why double doing the work. 

This approach does not differ much from my first attempt, where I'm currently working on and a push of changes comming in the next hours anyway - which changes codebase and making it more clean (and therefor going a little more in the direction of this approach). I'm also removed already the handling with the installed.json, making it simpler. I would suggest to review my change later when pushed instead of putting hours of work into something which is already worked on.

Just as first impressions:

I would not introduce here `EXT:` for referencing extension packages. My approach targeted to the direction to use the composer package name mainly, with the extension-key as fallback as intermediate solution/transition, not considered if removed later or not. Extension key fallback with introducing a new prefix (already known in the TYPO3 world). At a later point, functional and acceptance testcase instances should be build as composer installation, and therefore it make more sense to transition to the composer package name instead. We should move to composer standards and not introding a next extra level in the TF of a special key. 
Thinking about it - to have it expressive that a extension key is meant, maybe okay. However, we need a fallback to determine it anyway for the transition period. So not sure if it is worth to etablish this here. We should more enforce the transition to using the real composer name instead.

This implementation does not deal with the monorepo (coredev) structure, which has a root composer.json but not requiring the system extensions - and therefore missing these information then.

Some extensions out there uses `typo3conf/ext/extension-name` to eventually reference installed extensions which here is not handled by the implementation than, and failing
if root installation does not have it there (or using my bridge plugin) for cms-composer-installers 4RC1/5.

I like the class naming - something we could change in my approach then.

Maybe wait until my push later and decide if there is a need for doing this double approach. And finishing the missing parts.

Basically, both aproaches are the same except the "EXT:" prefix thing. Which I would not do from my point of view.

DavidBruchmann:
Yes I agree. I worked on it before I verified your PR, so I tested it already too, and that was the reason that it exists grossly in 2 versions now. Before moving to something else I thought the best is to push it first.

sbuerk:
> Main installation and acceptance tests never work yet, due to other reasons).

Don't get this. What do you mean exactly with this ? 

> Yes I agree. I worked on it before I verified your PR, so I tested it already too, and that was the reason that it exists grossly in 2 versions now. Before moving to something else I thought the best is to push it first.

My pull-request are quite some month old - so not really new. But to be honest, they have been a first POC implementation. Already had it on the list to redo and finish them, which
I was working on this weekend. Sadly, I did not push in-between. Pushed my solution for TF-main and TF-7 along with linked core test (main) and tests in several extensions (pull-requests) as evaluation.

Maybe you want to re-check / review my solution. I would say, that it is a little bit more complete, but I may be wrong. E.g. if tests have `typo3conf/ext/extension-key` configured for test extension to load, my solution would properly symlink that too. TBH not tested your solution, but from scrolling through the code I could not see something like that.

Except the `EXT:` syntax for package name selection, which I would avoid anyway. And having `vendor/package-name`, `extension-key` AND `EXT:extension-key` is kind of to much options. Rather explained that already.

I load the system extensions into the package informations for the mono-repo, as this is not done directly. However, in my solution it is symlinked than using the package information, versus your solution where in monorepo core system extensions are linked as before.

DavidBruchmann:
> > Main installation and acceptance tests never work yet, due to other reasons).
>
> Don't get this. What do you mean exactly with this ?

Thats related to the linked https://github.com/DavidBruchmann/t3docs-screenshots/, not to TF.

I'll have a look at your PR and will test it.
I'll test it first with the linked project.

Testing monorepo, I'd need a link to some ddev / docker project, I'm not so firm with it.

Let me know if you agree if I make some changes perhaps concerning naming or structure. Of course it depends on the detail and a yes wouldn't be a confirmation for everything.

sbuerk:
> > > Main installation and acceptance tests never work yet, due to other reasons).
> > 
> > 
> > Don't get this. What do you mean exactly with this ?
> 
> Thats related to the linked https://github.com/DavidBruchmann/t3docs-screenshots/, not to TF.

I see. So no TF/core issue.
 
> I'll have a look at your PR and will test it. I'll test it first with the linked project.
> Testing monorepo, I'd need a link to some ddev / docker project, I'm not so firm with it.

You don't need to test it in the core. We will do that and it's already backed by core ci (my PR, not yours). Having as much testing with extensions/projects out there would be nice.

I do not have any ddev for core monorepo as I do not use ddev at all. For tests, core uses testing framework and runtests, so in generell should not be much needed - simply clone monorepo and use `runTests.sh` for all things. But as already mentioned, no need for you to do that. We well do that with core merges and contributors the next days anyway.
 
> Let me know if you agree if I make some changes perhaps concerning naming or structure. Of course it depends on the detail and a yes wouldn't be a confirmation for everything.

Not sure what you want to change, but feel to suggest changes in my PR (as review/comment). Will then have a look and if they make sense incorporate them. You cannot change my PR anyway.

DavidBruchmann:
I get your point and agree some kind to this:

> Except the EXT: syntax for package name selection, which I would avoid anyway. And having vendor/package-name, extension-key AND EXT:extension-key is kind of to much options. Rather explained that already.

Nevertheless, it's not about extension key, but about outdated extension paths.

sbuerk:
> I get your point and agree some kind to this:
> 
> > Except the EXT: syntax for package name selection, which I would avoid anyway. And having vendor/package-name, extension-key AND EXT:extension-key is kind of to much options. Rather explained that already.
> 
> Nevertheless, it's not about extension key, but about outdated extension paths.

What do you mean by `it's not about extension key, but about outdated extension paths` ?

For example, following may be used out there:

```php
class SomeTest extends FunctionalTestCase
{
  protected array $testExtensionToLoad = [
    'typo3conf/ext/some_extension',
    'some_extensionkey',
    __DIR__ . '/Fixture/Extensions/test_extension',
  ];
}
```

Your move is, to introduce an additional prefix:

```php
class SomeTest extends FunctionalTestCase
{
  protected array $testExtensionToLoad = [
    'EXT:some_extension',     // instead of 'typo3conf/ext/some_extension'
    'EXT:some_extensionkey',  // instead of 'some_extensionkey'

    __DIR__ . '/Fixture/Extensions/test_extension',  // fallback like before
  ];
}
```

However, in my solution people do not need to change anything in there tests,
all old variants are working and symlinks are correctly created. In addition,
composer package name's can be used in my solution:

```php
class SomeTest extends FunctionalTestCase
{
  protected array $testExtensionToLoad = [
    'typo3conf/ext/some_extension', // finds the package and symlinks it,
                                    // even with cms-composer-installers 5
                                    // and installed in `<vendor-dir>/*/*`
    'some_extension',               // finds the package and symlinks it
    'some_extensionkey',            // finds the package and symlinks it
    'vendor/package-name',          // symlink package to as `extension-key` (NEW)

    __DIR__ . '/Fixture/Extensions/test_extension',  // fallback like before, with or
                                                     // without composer.json
  ];
}
```

Following is also possible now with my solution:

```php
class SomeTest extends FunctionalTestCase
{
  protected array $coreExtensionToLoad = [
    'core',              // simple extension-key
    'typo3/cms-backend', // package name works now too
  ];
}
```

Correct me, but I would say that my approach is here more cleaner. It works
direcelty, not enforcing to change from `extension-key` to `EXT:extension-key`.
That all combined is my reason, why I thing we should omit it.

Another benefit in my solution is, that only system extensions which have
been selected to be loaded (functional default's + $coreExtensionToLoad)
are now symlinked into a test instance path. Old behaviour was to symlink
the whole `typo3/sysext/` folder for each test instance only handling the
loaded system extensions for writing the test instance `PackageStates.php`.

In general, we want to build composer based test instances instead of legacy
onces in the future, albeit concreate implementation way is outstanding for
now. For that, dealing with composer package names makes more sense instead
of using extension-key only or enforcing a new prefix (`EXT:`) - which have to
be changed without automatic fallback.

Therefore, I would say the prefix is not needed with my solution and works
out-of-the box, and minimizing which extensions/projects needs to change to
get testing working again to a minimum (invoking unit/functional test with
phpunit / paths) instead of going through all test cases and have to change
a lot.

Further work may be, to enforce composer.json for test-fixture extensions
and detect/determine them automatically, registering them in the TF internal
ComposerPackageManager - so they can be referenced by the package-name/extension
name instead of a path. But I left that one out for now, to do only the needed
things to restore combatibility and working state.

I did not used your approach, and tested it (especially against the core). But
taking the experience and findings for my approach into account, I would say
that your solution misses a crucial point for the mono-repository - and I guess
it would fail (or using the old file fallback instead of composer information and
therefore slightly untedected).

In the mono-repository, the root `composer.json` contains replace configurations
for the system extension, and core extensions from `typo3/sysext/*` are not required.
That means, that in your solution the informations from composer installed version
are reported incorrectly. It would report the root composer.json and path for all
core extension package names. So that part is broken. In my solution, I added a 
mono-repo detection and load the `typo3/sysext/*/composer.json` informations manually,
but only for mono-repo. Otherwise using information from composer installed versions.

So, the `EXT:` thing is the really unique thing in your approach, which I would like
to avoid as explained above (because it is not really needed). Otherwise, my solution
is the more complete approach, tackling cases you not have detected yet or a solution
integrated.

Your implementation does not take local path packages into account, which the reported
`install-path` may not be the real source path. That means, that your solution would
symlink from the symlinked path in `vendor-dir`. My solution symlinks from the real
origin source to the test instance, avoiding nested symlinks at least by one level.

Thanks for your approach. Basically a nice approach code-wise, albeit the one or other
thing needed to be adjusted. But as explained detailed above, missing some cases.
I know that there have been quite some time put into it (I did the same), but I still
would go more with my solution from the implementation point-of-view - but we could
talk about namings.

I would really encourage you to test out my approach for your use-case. It should work,
at least until you have a edge-case I'm not aware off (which would not be handled by your
solution too).

lolli42:
Hey.

Stefan and me are not a huge fan of this PR: It adds quite a bit of code to achieve the issue, and we'd like to *not* introduce the `EXT:` syntax at all. *if* we go with something like this, we'd rather use composer package names instead, which we'll sort out when functional tests are refactored to create composer instances.

As such, this PR won't make it, and I hope it's ok to close this PR.

DavidBruchmann:
no problem.
actually I wait already long time that the functionality is merged, and I tested Stefan's PR, that's working similar and is doing it's job.

sbuerk:
> no problem. actually I wait already long time that the functionality is merged, and I tested Stefan's PR, that's working similar and is doing it's job.

Yeah - the PR was a first approach, but due priorities and other stuff got delayed.
It's in now merged and tf main (8.0.0) and 7 (7.0.0) tagged.

PR #463, opened 2023-04-26
[BUGFIX][v7] Explicitly require symfony/filesystem

eliashaeussler:
With c34df68, usage of the Symfony Filesystem's `Path` class was introduced. As this class is available since symfony/filesystem 5.4.0 only (see https://github.com/symfony/filesystem/commit/d5ca5f7080a759dcb65300a55d5ee42f1ee7ef59), it is essential to explicitly define this dependency with an appropriate version constraint. Otherwise, extensions running tests with TF and lowest dependencies may fail with TYPO3 < 11.5.5 because prior to this version, symfony/filesystem ^5.3 was required and therefore v5.3.x of the library may be installed (see https://github.com/TYPO3/typo3/commit/445f60667dc4db2da0d42c8b2faab8c67d5bd396).

See https://github.com/CPS-IT/handlebars/actions/runs/4802256934/jobs/8545498227 for an example of a failed test run.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
     "guzzlehttp/psr7": "^2.4.3",
     "phpunit/phpunit": "^9.5.25 || ^10.1",
     "psr/container": "^1.1.0 || ^2.0.0",
+    "symfony/filesystem": "^5.4 || ^6.0",
     "typo3/cms-backend": "11.*.*@dev || 12.*.*@dev",
     "typo3/cms-core": "11.*.*@dev || 12.*.*@dev",
     "typo3/cms-extbase": "11.*.*@dev || 12.*.*@dev",


eliashaeussler:
It seems like Composer now downgrades a lot of packages (e.g. TYPO3 v11.2.0) when installing dependencies to run tests in CI. I actually don't know why this happens 🤷. Temporarily changing TYPO3 version constraints to `11.5.*@dev || 12.*.*@dev` resolve this issue, but that's of course not a valid solution. Do you have any ideas?

sbuerk:
Not sure if we should add a direct require again.

Will think about this, but I guess I will replace the Path usage so this dependency is not needed. Leaving this PR open for now, but may close it later.

sbuerk:
> It seems like Composer now downgrades a lot of packages (e.g. TYPO3 v11.2.0) when installing dependencies to run tests in CI. I actually don't know why this happens shrug. Temporarily changing TYPO3 version constraints to `11.5.*@dev || 12.*.*@dev` resolve this issue, but that's of course not a valid solution. Do you have any ideas?

In the end, 11.0.x - 11.4.x are not supported, only >=11.5.0. Your project/extension should have that as bare minim version anyway. And I guess, most people does it that way, and therefore a "composer min install" check would have a sufficent symfony/filesystem package installed "anyway". This is the reason, why I did not stumpeled upon this issue, as only against core 11.5 has been tested.

Taken this into account, we do not have a "critical" time issue on here. But I will remove the usage for this package again, trying to avoid using the "Path" compononent. But I would say not that "urgent" atm, so this may be done with time (at the weekend). Gives time if further bugs/issues occures or are reported.

Would it be fine for you, to use the 11.5.x constrain in your project for now ? 

I have merged your other pull-request, which fixed a urgend bug. So we could tag a patchlevel nearby ?

sbuerk:
> > In the end, 11.0.x - 11.4.x are not supported, only >=11.5.0.
> 
> I think we then also probably need to modify the dependencies in the TF `composer.json` to only allow Core >= 11.5. Currently all versions of 11.x are support in the V7 branch if I'm reading this correctly:
> 
> ```
> "typo3/cms-core": "11.*.*@dev || 12.*.*@dev",
> ```

We can do that, and I thought about that to. But than at lest a "^8.0.1" is the needed requirement in the project. However, not thought well about it, but relying on a dependency which may introduce headache in later dual core versions, I would get rid of it anyway again in some way. But done with time and not in a hurry.

eliashaeussler:
> Would it be fine for you, to use the 11.5.x constrain in your project for now ?

I already have this constraint in my extension, see https://github.com/CPS-IT/handlebars/blob/8405e4cd69c370a1a6ed7f6c2fc4652c436896f9/composer.json#L27.

But, as I said, the problem here is that `typo3/cms-core` < 11.5.5 (which is a possible installation candidate with TF) requires `symfony/filesystem` as `^5.3`, leading to installation of

- `typo3/cms-core` 11.5.0
- `typo3/testing-framework` 7.0.0
- `symfony/filesystem` 5.3.0

when testing against lowest possible versions. And in this case, the `Path` class is missing.

I *could* in my own extension increase the minimum installable version of `typo3/cms-core` to 11.5.5 which would avoid any errors and solve this issue in this special case. But imho it should be the responsibility of the library itself (in this case: TF) to declare *their* dependencies and not rely on dependencies by transitive dependencies.

lolli42:
I think we should see if Stefan is able to get rid of the path thingie that requires sf/filesystem. If so, I'd opt to not change TF req's further, since that should be enough?

lolli42:
btw: why is the CI run of this PR red? I don't get it?

oliverklee:
The PHPStan failure might be due to a different version of Doctrine DBAL (possibly due to the used PHP version). For this reason, I'd suggest running PHPStan only with one PHP version and one TYPO3 version in CI. (I usually take the highest supported PHP and TYPO3 version for this.)

eliashaeussler:
I just found there's `TYPO3\CMS\Core\Utility\PathUtility::getCanonicalPath()` in TYPO3 Core, which could be a useful replacement for Symfony's `Path::canonicalize()`. Just dropping this in in case it helps. 🙂

sbuerk:
Closing this in favour of #467 and #468, containing an alternative approach for this.

PR #467, opened 2023-05-05
main stefan 1

sbuerk:
- [TASK] Avoid direct "symfony/filesystem" dependency
- [TASK] Include `ext_emconf.php` data in `PackageInfo`
- [TASK] Add some basic `ComposerPackageManager` tests
- [TASK] Simplify `resolvePackageName` for `ComposerPackageManager`
- [TASK] Normalize guessed extension key from path/packages.
- [BUGFIX] Ensure test fixture extension can be found
- [TASK] Handle replaced packages properly
- [TASK] Use phpunit configuration template for own testing


sbuerk:
Validated with core-main (pre-merge + nightly): https://review.typo3.org/c/Packages/TYPO3.CMS/+/78954

helhum:
> [TASK] Avoid direct "symfony/filesystem" dependency (altern

What is actually wrong with that dependency?

sbuerk:
> > [TASK] Avoid direct "symfony/filesystem" dependency (altern
> 
> What is actually wrong with adding that dependency?

Every direct dependency opens for possible version conflicts due multi-core support. We have to keep that up and have to deal with this. Therefore, we should avoid direct dependencies if possible and not absolutly needed.

E.g. in this case, we would require a newer version which was not there in TYPO3 11.0 (with TF7), and therefore changing dependencies and destroy composer min tests for project/extension users. If not absolutly required, we should keep TF support for sprint releases too.

And we do not need the complete package and power yet, so we can work around it.

lolli42 (approved):

PR #470, opened 2023-05-13
[TASK] Ensure root package extension is symlinked

sbuerk:
With https://github.com/TYPO3/testing-framework/pull/307 symlinking of the root package extensions is no
longer handled by the testing-framework composer script for
TYPO3 v11 based testing environments. This has been shifted
to the core.

TYPO3 v11 core implemention do no longer symlink root package
extensions to the extension folder, if no `Resources/Public/`
resources exists in the extension.

This change reverts the no-op check and let the root package
extension be symlinked again, in all circumstances.

This avoids this edge case.

**Note** TYPO3 v11 testing using `typor/cms-composer-installers`
version 4RC1 is still not supported. Switch to TF7+ for support.

Used command(s):

```shell
Build/Scripts/runTests.sh -s phpstanGenerateBaseline
```

Releases: 6
Related: https://github.com/TYPO3/testing-framework/commit/506837852eebbfe88c551058a5390323792f66f6

### DIFF
--- a/Build/phpstan/phpstan-baseline.neon
+++ b/Build/phpstan/phpstan-baseline.neon
@@ -11,11 +11,6 @@ parameters:
 			path: ../../Classes/Composer/ExtensionTestEnvironment.php
 
 		-
-			message: "#^Call to method getIO\\(\\) on an unknown class Composer\\\\Script\\\\Event\\.$#"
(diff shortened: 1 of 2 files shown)


helhum (approved):
Thanks!

PR #473, opened 2023-05-28
[BACKPORT][TASK] Configure frontend caches with NullBackend (#321)

sbuerk:
The patch configures frontend related caches to ese the
NullBackend in functional tests. This avoids creating corresponding
tables and reduces SELECT and INSERT database statements.
This results in a performance increase running frontend functional
tests, with core, this is around 5-10% when using postgres DMBS.

Late backport of e45ee97112096c5a902093cc9e2a4c07af1cb49e

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -27,7 +27,6 @@ use PHPUnit\Util\PHP\AbstractPhpProcess;
 use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use TYPO3\CMS\Core\Authentication\BackendUserAuthentication;
-use TYPO3\CMS\Core\Cache\Backend\NullBackend;
 use TYPO3\CMS\Core\Context\Context;
 use TYPO3\CMS\Core\Context\UserAspect;
 use TYPO3\CMS\Core\Core\Bootstrap;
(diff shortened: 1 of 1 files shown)


Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -380,8 +379,16 @@ protected function setUp(): void
 
             $localConfiguration['SYS']['trustedHostsPattern'] = '.*';
             $localConfiguration['SYS']['encryptionKey'] = 'i-am-not-a-secure-encryption-key';
-            $localConfiguration['SYS']['caching']['cacheConfigurations']['extbase_object']['backend'] = NullBackend::class;
             $localConfiguration['GFX']['processor'] = 'GraphicsMagick';
+            // Set cache backends to null backend instead of database backend let us save time for creating
+            // database schema for it and reduces selects/inserts to the database for cache operations, which
+            // are generally not really needed for functional tests. Specific tests may restore this in if needed.
+            $localConfiguration['SYS']['caching']['cacheConfigurations']['extbase_object']['backend'] = 'TYPO3\\CMS\\Core\\Cache\\Backend\\NullBackend';
```
helhum, line 386: This cache is called `extbase` since TYPO3 v10

helhum (changes requested):
Looks good, except Extbase cache

Classes/Core/Functional/FunctionalTestCase.php
sbuerk: Setting the "extbase" cache to a nullbackend has no benefit for the core itself.
Therefore, I decided to remove setting the invalid cache to NullBackend.

Extensions needing this or having a benefit of this can and should provide the
NullBackend override by the `configurationToUseInTestInstance = []` property.

Would this be fine for you @helhum ?

lolli42 (approved):

Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -380,8 +379,15 @@ protected function setUp(): void
 
             $localConfiguration['SYS']['trustedHostsPattern'] = '.*';
             $localConfiguration['SYS']['encryptionKey'] = 'i-am-not-a-secure-encryption-key';
-            $localConfiguration['SYS']['caching']['cacheConfigurations']['extbase_object']['backend'] = NullBackend::class;
             $localConfiguration['GFX']['processor'] = 'GraphicsMagick';
+            // Set cache backends to null backend instead of database backend let us save time for creating
+            // database schema for it and reduces selects/inserts to the database for cache operations, which
+            // are generally not really needed for functional tests. Specific tests may restore this in if needed.
+            $localConfiguration['SYS']['caching']['cacheConfigurations']['hash']['backend'] = 'TYPO3\\CMS\\Core\\Cache\\Backend\\NullBackend';
+            $localConfiguration['SYS']['caching']['cacheConfigurations']['imagesizes']['backend'] = 'TYPO3\\CMS\\Core\\Cache\\Backend\\NullBackend';
```
helhum, line 387: what is the reason now to not use the `NullBackend::class` notation here?

sbuerk: Because the original commit from main was done that way, and this is just a late backport.

PR #479, opened 2023-07-16
[BUGFIX] Harden XPath query to limit search for page tree node

eliashaeussler:
Even if page tree nodes are queried from a given context node, the currently used XPath query searches for **all elements in the document instead**. That's because of the `//` syntax, which searches in the whole document, regardless of which context is given. This may lead to possible test failures in case the queried node text appears **multiple times in the document**, for example when trying to access the root page tree node whose text appears twice on the page – next to the TYPO3 logo in the topbar and as root node in the page tree. With the current implementation, the XPath query would match the text in the topbar instead of the text of the root node.

In order to avoid such behaviors and actually limit the XPath query to the given context, it is now prefixed with a dot (`.`). In addition, only following siblings of the current node are included since page tree nodes are presented as flat node structure. This way, the current context is now actually taken into account.

As an additional fix, property accesses of static properties in `AbstractPageTree` is now changed from `self::` to `static::` to properly respect overridden properties in extended classes. This is especially necessary for the overridden properties in TYPO3 core's [`FileTree`](https://github.com/TYPO3/typo3/blob/v12.4.3/typo3/sysext/core/Tests/Acceptance/Support/Helper/FileTree.php#L26-L29).

Patch in TYPO3 core: https://review.typo3.org/c/Packages/TYPO3.CMS/+/80039

:bulb: If this PR will be merged, I'll update the TYPO3 core patch accordingly.

### DIFF
--- a/Classes/Core/Acceptance/Helper/AbstractPageTree.php
+++ b/Classes/Core/Acceptance/Helper/AbstractPageTree.php
@@ -52,7 +52,7 @@ abstract class AbstractPageTree
         foreach ($path as $pageName) {
             $context = $this->ensureTreeNodeIsOpen($pageName, $context);
         }
-        $context->findElement(\Facebook\WebDriver\WebDriverBy::cssSelector(self::$treeItemAnchorSelector))->click();
+        $context->findElement(\Facebook\WebDriver\WebDriverBy::cssSelector(static::$treeItemAnchorSelector))->click();
     }
 
     /**
@@ -65,7 +65,7 @@ abstract class AbstractPageTree
         $I = $this->tester;
         $I->switchToIFrame();
         return $I->executeInSelenium(function (\Facebook\WebDriver\Remote\RemoteWebDriver $webdriver) {
-            return $webdriver->findElement(\Facebook\WebDriver\WebDriverBy::cssSelector(self::$pageTreeSelector));
+            return $webdriver->findElement(\Facebook\WebDriver\WebDriverBy::cssSelector(static::$pageTreeSelector));
         });
     }
 
@@ -79,7 +79,7 @@ abstract class AbstractPageTree
     protected function ensureTreeNodeIsOpen(string $nodeText, RemoteWebElement $context)
     {
         $I = $this->tester;
-        $I->see($nodeText, self::$treeItemSelector);
+        $I->see($nodeText, static::$treeItemSelector);
 
         /** @var RemoteWebElement $context */
         $context = $I->executeInSelenium(function () use (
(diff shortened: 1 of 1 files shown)


oliverklee:
> As an additional fix, property accesses of static properties in AbstractPageTree is now changed from self:: to static:: to properly respect overridden properties in extended classes. This is especially necessary for the overridden properties in TYPO3 core's [FileTree](https://github.com/TYPO3/typo3/blob/v12.4.3/typo3/sysext/core/Tests/Acceptance/Support/Helper/FileTree.php#L26-L29).

I propose we move this to a separate change (following the "only one kind of change per PR" rule).

eliashaeussler:
> > As an additional fix, property accesses of static properties in AbstractPageTree is now changed from self:: to static:: to properly respect overridden properties in extended classes. This is especially necessary for the overridden properties in TYPO3 core's [FileTree](https://github.com/TYPO3/typo3/blob/v12.4.3/typo3/sysext/core/Tests/Acceptance/Support/Helper/FileTree.php#L26-L29).
> 
> I propose we move this to a separate change (following the "only one kind of change per PR" rule).

Sure, we can do that.

The reason why  I combined both changes into one PR is because the change of the XPath query makes tests in TYPO3 core fail due to the fact those static properties are not accessed correctly. This additional change was necessary to make tests in https://review.typo3.org/c/Packages/TYPO3.CMS/+/80039 green again 😉

But I'm also totally fine with splitting into two separate PRs.

sbuerk:
@eliashaeussler Don't act for now !!!

In TF we usually do not merge it as merge-commits, so as long as we
have dedicated commits for each thing it may be okay. Not qualified
reviewed it yet, will do this later.

The comment from @oliverklee in general is valid and a good advice
generally (makes it possible to merge one thing and eventually
continue on the rest).

In the worst case we (I) would cherry-pick the commits anyway.

Only point on quick view, it would be nice to have one or to sentence
in the commit message for 9f470629af4c246571c113b600e278eabdca7c88.

eliashaeussler:
@sbuerk Do you need anything else from me to continue here? :)

lolli42:
Ok. We now extracted the 'static' change with https://github.com/TYPO3/testing-framework/pull/625 to an own PR. Core tests stay green with this one for now, so we can reduce the PR a bit already without a core cross-dependency.

lolli42:
Ok, seems we're done here:

See https://github.com/TYPO3/testing-framework/pull/625
See https://github.com/TYPO3/testing-framework/pull/626

We cherry-picked the two (well split) commits to new PR's, of course keeping your authorship!
As such, we can close this PR now.
Sorry for the massive delay on this one, thes delay is not what we aim for.
Your two changes are done now, and we think it's a good improvement.
Thank you!

eliashaeussler:
@lolli42 Thanks for taking care!

PR #490, opened 2023-08-09
[TASK] Make all `in_array` calls strict

oliverklee:
This avoids false positives in cases where PHP would cast values.

Releases: main, 7, 6


lolli42:
no TF v7 and no v6, but i'll now run this through main und v12 core to see if it's fine.

lolli42 (approved):

PR #494, opened 2023-08-25
[TASK] Throw an exception if a BE user cannot be found in the DB

oliverklee:
This avoids passing `null` to a method that requires an `array`.

It also helps debugs tests that use an incorrect user UID for the session or that uses incorrect data in the DB.

This fix also is required to make more methods in `GeneralUtility` use native type declarations.

Releases: main


oliverklee:
Now running the core tests with this change:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/80693

lolli42 (approved):

PR #495, opened 2023-08-28
[TASK] Deprecate `getAccessibleMockForAbstractClass`

oliverklee:
This method uses `getMockForAbstractClass()` under the hood, which in turn has been soft-deprecated in PHPUnit 10.1 and will issue deprecation warnings in PHPUnit 11 and up:

https://github.com/sebastianbergmann/phpunit/issues/5241

Releases: main


oliverklee:
Stefan, Lolli, would this also make sense for version 7?

lolli42:
I think v8 / main TF should be sufficient?! Not using getMockForAbstractClass() is mainly a task ext authors may/should have to deal with with core 13 and above. since TF 8 will be 'core v12 & v13', we should probably note deprecation of it in TF 8 only. Is that fine, or did I mix up anything?

lolli42 (approved):

PR #513, opened 2023-10-20
[TASK] Replace `\PDO::PARAM_*` constants with `Connection::PARAM_*`

sbuerk:
This change replaces the `\PDO::PARAM_*` constants with the TYPO3
`\TYPO3\CMS\Core\Database\Connection::PARAM_*` alternatives. This
aligns the testing-framework with the core and the communicated
best-practice.

The direct dependency to `ext-pdo` is removed right away.

Note: `doctrine/dbal 4+` dropps support for the \PDO integer constants.
       Using the `Connection` constants allows us to change the mapping
       in that class and consuming code, like extensions or this package
       automaticlly uses the correct enum class on upgrade.

Used command(s):

```shell
composer remove ext-pdo
```

Releases: main, 7

### DIFF
--- a/Classes/Core/Functional/Framework/DataHandling/ActionService.php
+++ b/Classes/Core/Functional/Framework/DataHandling/ActionService.php
@@ -17,6 +17,7 @@ namespace TYPO3\TestingFramework\Core\Functional\Framework\DataHandling;
  */
 
 use TYPO3\CMS\Core\Authentication\BackendUserAuthentication;
+use TYPO3\CMS\Core\Database\Connection;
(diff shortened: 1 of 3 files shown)


oliverklee (approved):

oliverklee:
Should we also backport this to V7 to have that branch also follow our current best practices?

brotkrueml (approved):

oliverklee (changes requested):
We then can also drop the `ext-pdo` requirement from `composer.json`.

oliverklee (approved):

lolli42 (approved):

PR #524, opened 2023-12-01
[FEATURE] Add FakeEventDispatcher inspired by Laravel

simonschaufi:
See https://laravel.com/docs/10.x/events#testing for inspiration source and https://github.com/laravel/framework/blob/10.x/src/Illuminate/Support/Testing/Fakes/EventFake.php for the original source code.

How to use:

```php
use TYPO3\CMS\Core\EventDispatcher\ListenerProvider;

$fakeEventDispatcher = new FakeEventDispatcher(new ListenerProvider($this->getContainer()));

// $this->subject is an Extbase Controller
$this->subject->injectEventDispatcher($fakeEventDispatcher);

// run code which should dispatch an event (dispatch controller)

$fakeEventDispatcher->assertDispatched(MyEvent::class);
```

### DIFF
--- a/Classes/Core/Functional/Framework/Fakes/FakeEventDispatcher.php
+++ b/Classes/Core/Functional/Framework/Fakes/FakeEventDispatcher.php
@@ -1,0 +1,140 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TYPO3\TestingFramework\Core\Functional\Framework\Fakes;
+
+use PHPUnit\Framework\Assert as PHPUnit;
+use Psr\EventDispatcher\EventDispatcherInterface;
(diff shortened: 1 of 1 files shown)


lolli42:
Hey.

We decided to not go with this PR:
* In general, we want rather less than more helper classes in TF and only add things that are really very much needed by tests.
* Testing that events are actually dispatched by SUT often gives only very little value: dispatching an event is codewise no-brainish, there is little that can go wrong, having a test does not add crucial value.
* Devs who still want to check if events have been dispatched can easily do that with casual phpunit mocks and expectations. Having a helper class for this would add another "special handling" that can be solved with base phpunit API instead.

PR #528, opened 2023-12-28
[TASK] Respect `#[WithoutErrorHandler]` for error handler integrity check

sbuerk:
PHPUnit changed the error handler behaviour and introduced the PHP
attribute `#[WithoutErrorHandler]` to omit the error handler for
test methods.

This breaks the dangling error handler check provided by the
`typo3/testing-framework` which assumes that the PHPUnit error
handler is always the only registered error handler.

This change modifies the `BaseTestCase` class

* to provide the borrowed `shouldErrorHandlerBeUsed()` method
  cloned from `TestRunner` with additional version guards.
* to align the dangling error handler check in the `tearDown()`
  method of the `BaseTestCase` class using the added method.

Releases: main, 7
Related: https://forge.typo3.org/issues/102724


sbuerk:
Closing pull-request. Consideret not to backport this for now.

PR #529, opened 2024-01-10
[TASK] Ensure Doctrine DBAL 3 and 4 dual support

sbuerk:
* [1] Doctrine DBAL 4 removed `SqliteManager->createDatabase()`
  and `SqliteManager->dropDatabase()`. The engine is createing
  the SQLite file on connect, but do not remove it on close.
  The `Testbase->setUpTestDatabase()` code is modified to delete
  the sqlite file directly, and let the engine handle the sqlite
  file creation on connect.

* [2] Doctrine DBAL 4 changed the casing of some classes, for
  exmaple `SqlitePlatform` to `SQLitePlatform`, which is not an
  issue for code execution due to the case-insensitive nature of
  PHP - but static code analyser tools like PHPStan fall on their
  feed. Therefore, we use minor wrapping methods instead of php
  interfaceof checks directly to mitigate this and avoid jumping
  PHPStan baseline entry until Doctrine DBAL 4 is the minimal to
  support version.

  Note: The Doctrine DBAL 4 namespace casing is used.

[1] https://github.com/doctrine/dbal/blob/3.7.x/UPGRADE.md#deprecated-sqliteschemamanagercreatedatabase-and-dropdatabase-methods
[2] https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md#bc-break-renamed-sqlite-platform-classes

Releases: main

### DIFF
--- a/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseSnapshot.php
+++ b/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseSnapshot.php
@@ -18,7 +18,6 @@ declare(strict_types=1);
 namespace TYPO3\TestingFramework\Core\Functional\Framework\DataHandling\Snapshot;
 
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Platforms\SqlitePlatform as DoctrineSQLitePlatform;
 
 /**
  * Implement the database snapshot and callback logic.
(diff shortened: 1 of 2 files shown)


sbuerk:
Note: Tested with main + Doctrine DBAL 3 -> https://review.typo3.org/c/Packages/TYPO3.CMS/+/82428

bmack (approved):

PR #530, opened 2024-01-11
[TASK] Avoid json encoding of `ParameterType`

sbuerk:
Doctrine DBAL 4 transform the `ParameterType`
from class to a native enum, which is a non
baked enum which prevents serialization of
the enum, for example using `json_encode()`.

The `DatabaseAccessor` used for the function
test `withSnapshot()` feature encodes column
information of the snapshotted data as json,
badly breaking down with `Doctrine DBAL 4`.

`TYPO3\CMS\Core\Database\Connection` has been
modified to determine the parameter types for
`insert()`, `update()` and similar methods
based on the database structure if the types
are not passed.

This change modifes the column preparation
to avoid the enum type and instead save the
doctrine type name as string. This is only
done, if `ParameterType` is a real enum and
therefore only with `Doctrine DBAL 4`.

The adjustment on the `Connection` is added
to the `Doctrine DBAL 4 Upgrade` change and
is therefore a tight coupled bridge.

[1] https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md#bc-break-converted-enum-like-classes-to-enums
[2] https://review.typo3.org/c/Packages/TYPO3.CMS/+/76553

Releases: main

### DIFF
--- a/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseAccessor.php
+++ b/Classes/Core/Functional/Framework/DataHandling/Snapshot/DatabaseAccessor.php
@@ -17,7 +17,10 @@ declare(strict_types=1);
 
 namespace TYPO3\TestingFramework\Core\Functional\Framework\DataHandling\Snapshot;
 
+use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Schema\Table;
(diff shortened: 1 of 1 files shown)


sbuerk:
Testing with https://review.typo3.org/c/Packages/TYPO3.CMS/+/82428

PR #531, opened 2024-01-16
[TASK] Add dummy coverage configuration for PHPUnit

oliverklee:
Newer PHPUnit versions require the configuration files to have a `coverage` node (which can also be empty, but it needs to exist). Otherwise, PHPUnit will complain about the configuration file being outdated.

### DIFF
--- a/Resources/Core/Build/FunctionalTests.xml
+++ b/Resources/Core/Build/FunctionalTests.xml
@@ -31,6 +31,7 @@
          failOnWarning="true"
          requireCoverageMetadata="false"
 >
+    <coverage/>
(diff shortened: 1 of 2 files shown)


sbuerk:
why should be a empty block be mantandory ? Maybe a bug in PHPUnit ? 

Do you have a link/evidence regarding this that this is "intentended" by phpunit ?

oliverklee:
Ah, it seems that this configuration file node was only used up to PHPUnit 9 and got removed in PHPUnit 10:
https://github.com/sebastianbergmann/phpunit/issues/5294

PR #532, opened 2024-01-18
[TASK] Update to PHPUnit 10.5

oliverklee:
The Core (`main` and `12.4`) already is on 10.5.


lolli42:
This is obsolete in main which has been raised to 11 already.
For 8: I actually tend to not raise this anymore to not add potential additional trouble with any consumer. There is no strict need to upgrade, either. I hope this is fine with you.

PR #533, opened 2024-01-22
[TASK] Use /index.php as application entry point

bnf:
/typo3/index.php is deprecated and must not longer be used.

See https://forge.typo3.org/issues/87889 and
https://review.typo3.org/c/Packages/TYPO3.CMS/+/74366

Releases: main

### DIFF
--- a/Classes/Core/Acceptance/Helper/Login.php
+++ b/Classes/Core/Acceptance/Helper/Login.php
@@ -39,7 +39,7 @@ class Login extends Module
     ];
 
     /**
-     * Set a backend user session cookie and load the backend index.php.
+     * Set a backend user session cookie and load the backend entrypoint.
      *
      * Use this action to change the backend user and avoid switching between users in the backend module
      * "Backend Users" as this will change the user session ID and make it useless for subsequent calls of this action.
@@ -55,21 +55,21 @@ class Login extends Module
 
         $hasSession = $this->_loadSession();
         if ($hasSession && $newUserSessionId !== '' && $newUserSessionId !== $this->getUserSessionId()) {
-            $webDriver->amOnPage('/typo3/index.php');
+            $webDriver->amOnPage('/typo3');
             $webDriver->wait($waitTime);
             $this->_deleteSession();
             $hasSession = false;
         }
 
         if (!$hasSession) {
-            $webDriver->amOnPage('/typo3/index.php');
+            $webDriver->amOnPage('/typo3');
             $webDriver->wait($waitTime);
             $webDriver->waitForElement('body[data-typo3-login-ready]');
             $this->_createSession($newUserSessionId);
         }
 
         // Reload the page to have a logged in backend.
-        $webDriver->amOnPage('/typo3/index.php');
+        $webDriver->amOnPage('/typo3');
         $webDriver->wait($waitTime);
 
         // Ensure main content frame is fully loaded, otherwise there are load-race-conditions ..
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -30,6 +30,7 @@ use TYPO3\CMS\Core\Core\ClassLoadingInformation;
 use TYPO3\CMS\Core\Core\Environment;
 use TYPO3\CMS\Core\Database\Connection;
 use TYPO3\CMS\Core\Database\ConnectionPool;
+use TYPO3\CMS\Core\Http\Application as CoreHttpApplication;
 use TYPO3\CMS\Core\Http\NormalizedParams;
 use TYPO3\CMS\Core\Http\ServerRequest;
 use TYPO3\CMS\Core\Http\Stream;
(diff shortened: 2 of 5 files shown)


sbuerk:
quick view/first impression:

we need to make this conditional based to ensure working with v12 (state before change) and v13 (with changed entrypoints) - or start breaking the 1 TF version 2 core version support away, which I would try to avoid as this will get a huge cry in the community and extension developers.

bnf:
Hmpf....

Current state is a huge mess for us developers and results in ugly code all over the place (For example all that outdated code in `Resources/Core/Build/Configuration/JSUnit` where our only chance of properly improving the core – adding importmaps – was to ignore that outdated and non-changeable configuration in Testing Framework and craft an own jsunit setup in TYPO3).

Providing *API* compatibility for two versions might be fine, but that doesn't mean one TestingFramework version needs to support multiple TYPO3 versions. Extension authors can simply require-dev `"typo3/testing-framework": "^8 || ^9"`.

Note: I also want to refactor other stuff like the initialization with `SystemEnvironmentBuilder::run(…, SystemEnvironmentBuilder::REQUESTTYPE_CLI)` later on. Reason: That initialization is wrong (tests run with HTTP headers but initialized as CLI) and means we do a bunch of tests based on wrong assumptions (and they fail badly when that wrong assumption is corrected, often because the test data is broken as well. I'd like to get that fixed but I guess that won't be doable with a TF that supports v12 and v13.

So please enlighten me on why we can't simply release TF v9 with the same *API* as v8?
Releasing v9 (or even better name it v13) will also help to simply merge this repo into core as we did with EXT:styleguide and use the same deprecation policy as for our core code.

bnf:
> we need to make this conditional based to ensure working with v12 (state before change) and v13 (with changed entrypoints) 

I've now added such conditions for v12 compat and tested on current 12.4 and main via:

* https://review.typo3.org/c/Packages/TYPO3.CMS/+/82595
* https://review.typo3.org/c/Packages/TYPO3.CMS/+/82596

I still think we should do the split versions with focus on merging TF into core, but I'll discuss that when we've got more time for such things.

bmack:
I'm fine with the change as is. @sbuerk ?

lolli42:
tl;dr: Patch looks good, especially with https://review.typo3.org/c/Packages/TYPO3.CMS/+/82595 being green.

About versioning:
Yes, we can branch-off v8 and start a v9 in main now. But we'll need the patch in v8 as well since v8 needs to continue keeping core v13 compat. That's our current deal: use v8 to run extension tests with both v12 and v13 core. When we branch-off, main can get the b/w compat todo's resolved.

The reason for TF supporting two core versions: This guarantees API compatibility for extensions, which is kinda convenient for extension authors. It also forces TF maintainers to think about compat layers. We basically have a triangle between extension<->TF-api<->core-api here. In case of this patch TF<->core is affected, which forces us to have TF compat at this point to keep extension<->TF clean.

Note I'd love to see a TF core-monorepo merge as well, but this needs a bit of additional infrastructure since monorepo currently can't host casual composer packages. When we merge at some point, we could review the compat problematic along the way.

Classes/Core/Acceptance/Helper/Login.php
```diff
@@ -55,21 +55,21 @@ public function useExistingSession(string $role, float|int $waitTime = 0.5)
 
         $hasSession = $this->_loadSession();
         if ($hasSession && $newUserSessionId !== '' && $newUserSessionId !== $this->getUserSessionId()) {
-            $webDriver->amOnPage('/typo3/index.php');
+            $webDriver->amOnPage('/typo3');
```
lolli42, line 58: those are ok due to v12 / web server redirects.

Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -496,15 +497,20 @@ private function createServerRequest(string $url, string $method = 'GET'): Serve
     {
         $requestUrlParts = parse_url($url);
         $docRoot = $this->instancePath;
+
+        // @todo: Remove when dropping support for v12
+        $hasConsolidatedHttpEntryPoint = class_exists(CoreHttpApplication::class);
```
lolli42, line 502: that's a simple and straight b/w compat detection strategy. fine with me. and it's smart since v13 core without the entry-point patch does not fail (which would happen when going for a typo3Version check).

Classes/Core/Functional/FunctionalTestCase.php
```diff
@@ -944,19 +950,23 @@ private function retrieveFrontendSubRequestResult(
...
-            Environment::getProjectPath(),
-            Environment::getPublicPath(),
-            Environment::getVarPath(),
-            Environment::getConfigPath(),
-            Environment::getPublicPath() . '/index.php',
-            Environment::isWindows() ? 'WINDOWS' : 'UNIX'
-        );
+        // @todo: Remove when dropping support for v12
+        $hasConsolidatedHttpEntryPoint = class_exists(CoreHttpApplication::class);
+        if (!$hasConsolidatedHttpEntryPoint) {
```
lolli42, line 955: So good to see this gone in v13!

lolli42 (approved):

sbuerk (approved):

bnf:
thx!

Issue #535, opened 2024-01-25
BackendUserHandler breaks workspace


DanielSiepmann:
### What are you trying to achieve?

A frontend request with enabled backend user.
As we extend admin panel and want to cover that with functional tests executing frontend requests.

### What do you get instead?

We receive a 404 due to `BackendUserHandler` class which breaks the expected workspace aspect.

### How to reproduce the issue?

Execute the following test:

```php
        $request = new InternalRequest();
        $request = $request->withPageId(1);
        $request = $request->withLanguageId($languageUid);
        $requestContext = new InternalRequestContext();
        $requestContext = $requestContext->withBackendUserId(1);
        $html = (string) $this->executeFrontendSubRequest($request, $requestContext)->getBody();
```

With the following database fixture:

```php
<?php

declare(strict_types=1);

return [
    'pages' => [
        [
            'uid' => 1,
            'pid' => 0,
            'title' => 'Page with uid 1 below 0',
            'deleted' => 0,
            'backend_layout' => 'pagets__21',
            // Necessary to allow user to access this page.
            'perms_everybody' => 1,
            'slug' => '/start',
        ],
    ],
    'be_users' => [
        [
            'uid' => 1,
            'pid' => 0,
            'tstamp' => 1366642540,
            'username' => 'editor',
            // The actual password is: password
            'password' => '$1$tCrlLajZ$C0sikFQQ3SWaFAZ1Me0Z/1',
            'admin' => 0,
            'disable' => 0,
            'starttime' => 0,
            'endtime' => 0,
            'options' => 0,
            'crdate' => 1366642540,
            'workspace_perms' => 1,
            'deleted' => 0,
            'TSconfig' => null,
            'lastlogin' => 1371033743,
            'workspace_id' => 0,
            // Necessary to allow user to access this page.
            'db_mountpoints' => 1,
            // Necessary to open admin panel.
            'uc' => serialize([
                'AdminPanel' => [
                    'display_top' => true,
                ],
            ]),
        ],
    ],
];
```

### Additional information you would like to provide?

It looks like https://github.com/TYPO3/testing-framework/blob/8.0.8/Resources/Core/Functional/Extensions/json_response/Classes/Middleware/BackendUserHandler.php doesn't properly set the workspace property until one explicitly provides it within the test. Leading to keeping the -99 fallback which in turn will lead to other follow up issues as the system status is wrong.

### Specify some data of the environment

- TYPO3 testing framework version: 8.0.8
[...]

PR #536, opened 2024-01-25
[BUGFIX] Do not end up in workspace and preview for default requests

DanielSiepmann:
Default requests won't have a workspace id.
That will lead to TYPO3 defaults -99 id.
That in turn will lead to being in preview mode once a backend user is provided within context.

We therefore cover the default and turn it into live workspace which could be the expected default.

Resolves: #535


oliverklee:
Should this only go into the `main` branch, or would it make sense to backport it to 7.x as well?

DanielSiepmann:
I'm fine with both. I personally need this in one project where we apply this as a patch.
I'd leave this up to the maintainer to decide.

Resources/Core/Functional/Extensions/json_response/Classes/Middleware/BackendUserHandler.php
```diff
@@ -55,6 +55,10 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
             // Init backend user if found in database
             $backendUser = GeneralUtility::makeInstance(FrontendBackendUserAuthentication::class);
             $backendUser->user = $row;
+            $backendUser->uc = isset($row['uc']) ? unserialize($row['uc']) : [];
+            $backendUser->initializeUserSessionManager();
+            $backendUser->fetchGroupData();
+            $workspaceId = $workspaceId ?? 0;
```
sbuerk, line 61: This literlly renders the condition on the next line use-less. Beside that, I'd say that we should go here more for using `$backendUser->setDefaultWorkspace()`, not fallback to `0` and only set the temporary workspace id if specified in the internal request context.

What do you think ?

DanielSiepmann: Ah sure, didn't notice that one.

Only setting if provided would not solve the issue. I would need to dig into it once more. But the issue arised by not providing a workspace id.

sbuerk: never mind - `setDefaultWorkspace()` is protected. However, the if condition should be removed and setTemporaryWorkspace() always called with the 0 fallback.

```php
        if ($row !== false) {
            // Init backend user if found in database
            $backendUser = GeneralUtility::makeInstance(FrontendBackendUserAuthentication::class);
            $backendUser->user = $row;
            $backendUser->uc = isset($row['uc']) ? unserialize($row['uc']) : [];
            $backendUser->initializeUserSessionManager();
            $backendUser->fetchGroupData();
            $backendUser->setTemporaryWorkspace($workspaceId ?? 0);
            $GLOBALS['BE_USER'] = $backendUser;
            $this->setBackendUserAspect(GeneralUtility::makeInstance(Context::class), $backendUser);
        }
```

DanielSiepmann: Sure, done.

sbuerk:
If I run full core functional testsuite (with your original change and also my adoption) i get 4 failures:

```
$ Build/Scripts/runTests.sh -s functional

1) TYPO3\CMS\Core\Tests\Functional\DataScenarios\Regular\WorkspacesPublish\ActionTest::hideContent
Failed asserting that response has records.
* Section "Default": Could not assert all values for "tt_content.header": Regular Element #2

/var/www/work/t3c/core-main/core-main/typo3/sysext/core/Tests/Functional/DataScenarios/Regular/WorkspacesPublish/ActionTest.php:105

2) TYPO3\CMS\Core\Tests\Functional\DataScenarios\Regular\WorkspacesPublish\ActionTest::hideContentAndMoveToDifferentPage
Failed asserting that response has records.
* Section "Default": Could not assert all values for "tt_content.header": Regular Element #2

/var/www/work/t3c/core-main/core-main/typo3/sysext/core/Tests/Functional/DataScenarios/Regular/WorkspacesPublish/ActionTest.php:133

3) TYPO3\CMS\Core\Tests\Functional\DataScenarios\Regular\WorkspacesPublishAll\ActionTest::hideContent
Failed asserting that response has records.
* Section "Default": Could not assert all values for "tt_content.header": Regular Element #2

/var/www/work/t3c/core-main/core-main/typo3/sysext/core/Tests/Functional/DataScenarios/Regular/WorkspacesPublishAll/ActionTest.php:100

4) TYPO3\CMS\Core\Tests\Functional\DataScenarios\Regular\WorkspacesPublishAll\ActionTest::hideContentAndMoveToDifferentPage
Failed asserting that response has records.
* Section "Default": Could not assert all values for "tt_content.header": Regular Element #2

/var/www/work/t3c/core-main/core-main/typo3/sysext/core/Tests/Functional/DataScenarios/Regular/WorkspacesPublishAll/ActionTest.php:128
```

so I guess we need more investigation there first. Will have a more detailed look later.

Issue #541, opened 2024-02-14
Add support for dependencies to TYPO3 Extensions without ext_emconf.php


DanielSiepmann:
### Does your feature request relate to a problem?

We removed all our `ext_emconf.php` files when updating from TYPO3 10 LTS to 12 LTS.
This leads to issues like:
```
UnexpectedValueException: The package "e2_ces" depends on "e2/e2-core" which is not present in the system.
```

### Describe the solution you would like

Provide native support for composer only projects with multiple TYPO3 extensions that declare their dependencies on each other (for loading order).

### Optionally, describe alternatives you considered

One could keep ext_emconf.php.

DanielSiepmann:
Probably related to https://github.com/TYPO3/testing-framework/issues/511 and https://github.com/TYPO3/testing-framework/issues/510.

websi:
This is really annoying. Every new release our test fail because this will not be fixed.

It would be really nice if we can use the newest version with extension without "ext_emconf.php".

All our extension has no "ext_emconf.php".

DanielSiepmann:
I already opened a PR which should implement the solution.
I got distracted and couldn't finish my investigation regarding acceptance tests. Feel free to step in and investigate the issue. I hope this is the last missing piece.

PR #542, opened 2024-02-14
[BUGFIX] Respect composer mode only extension in FunctionalTestCase

DanielSiepmann:
The `typo3/testing-framework` creates functional test instances
as "classic" mode instances, writing a `PackageStates.php` file
composed using composer information, provided and handled by the
internal `ComposerPackageManager` class. Extensions in the state
file are not sorted based on their dependencies and suggestions.

To mitigate this and having a correctly sorted extension state
file, the `PackageCollection` service has been introduced with
the goal to resort the extension state file after the initial
write to provide the correct sorted extension state.

The extension sorting within the state file is important, as the
TYPO3 Core uses this sorting to loop over extensions to collect
information from the extensions, for example TCA, TCAOverrides,
ext_localconf.php and other things. Package sorting is here very
important to allow extensions to reconfigure or change the stuff
from other extensions, which is guaranteed in normal "composer"
and "classic" mode instances.

For "classic" mode instances, only the `ext_emconf.php` file is
taken into account and "composer.json" as the source of thruth
for "composer" mode instances since TYPO3 v12, which made the
`ext_emconf.php` file obsolete for extensions only installed
with composer.

Many agencies removed the optional `ext_emconf.php` file for
project local path extensions like a sitepackage to remove
the maintenance burden, which is a valid case.

Sadly, the `PackageCollection` adopted from the TYPO3 Core
`PackageManager` did not reflected this case and failed for
extensions to properly sort only on extension dependencies
and keys due the mix of extension key and composer package
name handling. Extension depending on another extension
failed to be sorted correctly with following exception:

    UnexpectedValueException: The package "extension_key"
    depends on "composer/package-key" which is not present
    in the system.

This change modifies the `PackageCollection` implementation
to take only TYPO3 extension and system extension into account
for dependency resolving and sorting, using `ext_emconf.php`
depends/suggest information as first source and falling back
to `composer` require and suggestion information.

Resolves: https://github.com/TYPO3/testing-framework/issues/541
Releases: main, 8

### DIFF
--- a/Build/phpstan/phpstan.neon
+++ b/Build/phpstan/phpstan.neon
@@ -20,3 +20,5 @@ parameters:
     - ../../Classes/Core/Acceptance/*
     # Text fixtures extensions uses $_EXTKEY phpstan would be report as "might not defined"
     - ../../Tests/Unit/*/Fixtures/Extensions/*/ext_emconf.php
+    - ../../Tests/Unit/*/Fixtures/Packages/*/ext_emconf.php
+    - ../../Tests/Unit/Fixtures/Packages/*/ext_emconf.php
--- a/Classes/Composer/ComposerPackageManager.php
+++ b/Classes/Composer/ComposerPackageManager.php
@@ -61,6 +61,7 @@ final class ComposerPackageManager
 
     public function __construct()
     {
+        // @todo Remove this from the constructor.
         $this->build();
     }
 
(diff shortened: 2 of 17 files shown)


oliverklee:
Should this only go into the `main` branch, or would it make sense to backport it to 7.x as well?

DanielSiepmann:
Looks like there is somehow an issue with loading order within acceptance tests … I'll investiage.

liayn:
Hm... Feature?
I would rather call this a serious bugfix...

All of our internal extensions skip emconf (obviously) so this hits us on every upgrade from v11 to v12 of each extension. :-(

sbuerk:
on it ..

liayn:
@sbuerk  Thanks a bunch. I just wonder what we changed in Core v12 that this hits so hard? I mean this worked flawlessly - at least for us - with Core v11 and TF 7

sbuerk:
> @sbuerk Thanks a bunch. I just wonder what we changed in Core v12 that this hits so hard? I mean this worked flawlessly - at least for us - with Core v11 and TF 7

TF8 added the sorting of packages to ensure that extension in functional instances has the correct sorting, normally TYPO3 would take care of in a legacy instance. As the package state is written by tf after symlinking to it, the state file must be sorted. For that, a custom sorting thing has been implemented in TF to take care of that.

With TF7 / v11 or v12 you have to take care to mention extension in the correct order yourself.

For example, of you have a extension which requries to be executed after another extension. For example having suggest defined so your extension runs after for example EXT:dashboard to register widgets etc.

The `PackageCollection` just assums (use code which assums) that extension still have ext_emconf.php in the extension.

There are lot more details to this, but we need to address some things in this PackageCollection now.

sbuerk:
@DanielSiepmann Modified your commit/pull-request.

Note, testest the very same commit with https://github.com/TYPO3/testing-framework/pull/634 against TYPO3 Core with https://review.typo3.org/c/Packages/TYPO3.CMS/+/87253 and also
in a nightly.

Is the change okay for you ?

d-s-codappix:
Sure. I've replaced all `ext_emconf.php` files of our test fixture extensions with corresponding `composer.json` files and tested that state with the latest version of this PR. This version works fine for us in our v12 project with TF 8.

lolli42 (approved):

PR #543, opened 2024-02-26
[TASK] Do not hard require typo3/cms-install

lolli42:
TF itself has no hard dependency to typo3/cms-install.

> composer rem typo3/cms-install

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,8 +33,7 @@
     "typo3/cms-core": "12.*.*@dev || 13.*.*@dev",
     "typo3/cms-extbase": "12.*.*@dev || 13.*.*@dev",
     "typo3/cms-fluid": "12.*.*@dev || 13.*.*@dev",
-    "typo3/cms-frontend": "12.*.*@dev || 13.*.*@dev",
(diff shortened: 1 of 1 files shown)


sbuerk (approved):

mbrodala:
Actually there currently is a hard dependency, at least on TYPO3v12:

https://github.com/TYPO3/testing-framework/blob/8.1.0/Classes/Core/Testbase.php#L250

This currently lets functional tests fail:

https://github.com/pagemachine/typo3-formlog/actions/runs/9609757403/job/26504856080

> UnexpectedValueException: Source file (/.../typo3temp/var/tests/functional-3b7c664/typo3/sysext/install/Resources/Private/Php/install.php) was not found.

Also the check for `$coreExtensions` does not work here since `install` is always loaded:

https://github.com/TYPO3/testing-framework/blob/8.1.0/Classes/Core/Functional/FunctionalTestCase.php#L298-L301

So this should be reverted or further adjustments are necessary to make `install` also optional on TYPO3v12.

Of course, the workaround for now would be adding `typo3/cms-install` to `require-dev`.

PR #562, opened 2024-05-23
[FEATURE] Clean up properties in BaseTestCase tearDown and add option to show memory info

### LINKED ISSUE #561: Show memory lost/leaked per test and auto-fix memory leaks fot tests that have no proper tear-down
### Does your feature request relate to a problem?
A complete test-run needs huge amounts of memory.

Especially the TYPO3 functional tests:

```
$ ./Build/Scripts/runTests.sh -s functional -p 8.2 -d mariadb

............................................................. 7137 / 7862 ( 90%)
............................................................. 7198 / 7862 ( 91%)
............................................................. 7259 / 7862 ( 92%)
....................S.S...................................... 7320 / 7862 ( 93%)
............................................................. 7381 / 7862 ( 93%)
...........................................................PHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 299008 bytes) in /TYPO3.CMS/typo3/sysext/core/Classes/Cache/Backend/SimpleFileBackend.php on line 380

Fatal error: Allowed memory size of 2147483648 bytes exhausted (tried to allocate 299008 bytes) in /TYPO3.CMS/typo3/sysext/core/Classes/Cache/Backend/SimpleFileBackend.php on line 380

###########################################################################
Result of functional
Container runtime: docker
PHP: 8.2
DBMS: mariadb  version 10.4  driver mysqli
FAILURE
###########################################################################
```

### Describe the solution you would like

Add some automatic tear-down of test-properties to save memory on the complete test-runs.
And display the amount of 'leaked'/'lost' memory for each test-case.

Also see https://www.drupal.org/project/drupal/issues/3280882

### Optionally, describe alternatives you considered

The test-implementations itself initialize properties in `setUp` method and should use `tearDown` method to remove these properties again.

Unfortunately `tearDown` is not implemented in every test-implementation or not all the properties are accurately unset.

So maybe somehow mark tests that do not properly tear-down their properties. And then these tests can be fixed one by one.

rr-it:
This adds automatic tear-down of test-object properties to save memory on test-runs.
Garbage collection is triggered in `tearDownAfterClass` method to really free up memory.

Fixes #561

----

This introduces a new toggle via environment variable `TYPO3TESTINGFRAMEWORK_MEMORY_INFO` :
If  enabled:

- The amount of additional memory usage and total memory usage is shown after each test-case.
- In case a property was not properly unset by the test-case itself an exclamation mark '!' is added to the output.

This variable can be set e.g. in `FunctionalTests.xml`/`UnitTests.xml`:

```xml
<?xml version="1.0"?>
<phpunit>
    <php>
        <env name="TYPO3TESTINGFRAMEWORK_MEMORY_INFO" value="1"/>
    </php>
</phpunit>
```

Example output:

```
$ ./Build/Scripts/runTests.sh -s unit -p 8.2 -d mariadb
PHPUnit 11.0.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.19
Configuration: /TYPO3.CMS/Build/phpunit/UnitTests.xml

# TYPO3\CMS\Adminpanel\Tests\Unit\Utility\StateUtilityTest
...............
Memory: +0.02 MB / 205.83 MB

# TYPO3\CMS\Backend\Tests\Unit\CodeEditor\Registry\ModeRegistryTest
!.!.!.
Memory: +0.03 MB / 205.86 MB

# TYPO3\CMS\Backend\Tests\Unit\Configuration\BackendUserConfigurationTest
!!.!!.!!.!!.!!.!!.!!.
Memory: +0.56 MB / 206.42 MB
```

```
$ ./Build/Scripts/runTests.sh -s functional -p 8.2 -d mariadb
Using driver: mysqli
PHPUnit 11.0.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.19
Configuration: /TYPO3.CMS/Build/phpunit/FunctionalTests.xml

# TYPO3\CMS\Core\Tests\Functional\DataHandling\DataHandler\HookTest
....
Memory: +2.37 MB / 151.20 MB

# TYPO3\CMS\Core\Tests\Functional\DataHandling\DataHandler\PagePermissionTest
!!.!!.!!.
Memory: +8.76 MB / 159.96 MB

# TYPO3\CMS\Core\Tests\Functional\DataHandling\DataHandler\SecurityTest
!.!.!.!.!.!.!.!.!.!.!.!.  854 / 7862 ( 10%)
!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.!.
Memory: +101.25 MB / 252.45 MB

# TYPO3\CMS\Core\Tests\Functional\DataHandling\DataHandler\SelectCheckBoxTest
!.!.
Memory: +3.40 MB / 255.85 MB
```

### DIFF
--- a/Classes/Core/BaseTestCase.php
+++ b/Classes/Core/BaseTestCase.php
@@ -38,6 +38,15 @@ abstract class BaseTestCase extends TestCase
      */
     private ?int $backupErrorReporting = null;
 
+    public static function setUpBeforeClass():void
+    {
+        gc_collect_cycles();
+        $GLOBALS['memory_usage_start'] = memory_get_usage();
+        echo(PHP_EOL . '# ' . static::class . PHP_EOL);
+
+        parent::setUpBeforeClass();
+    }
+
     protected function setUp(): void
     {
         parent::setUp();
@@ -86,6 +95,35 @@ abstract class BaseTestCase extends TestCase
                 );
             }
         }
+
+        /*
+         * Unset properties introduced by the test-class itself
+         *
+         * All these properties should actually be unset by the test-class tearDown
+         * method.
+         * Unfortunately tearDown is not implemented in every test-class or the
+         * implemented tearDown does not accurately unset all properties.
+         *
+         * Note: Private properties are not found and also cannot be automatically
+         * cleaned up here.
+         */
+        $rc = new \ReflectionClass(static::class);
+        foreach ($rc->getProperties(\ReflectionProperty::IS_PUBLIC | \ReflectionProperty::IS_PROTECTED) as $property) {
+            if (!$property->isStatic() && $property->class === static::class && isset($this->{$property->name})) {
+                // Auto-cleanup:
+                unset($this->{$property->name});
+                echo '!';
+            }
+        }
+    }
+
+    public static function tearDownAfterClass():void
+    {
+        parent::tearDownAfterClass();
+
+        gc_collect_cycles();
+        echo(PHP_EOL . 'Memory leak: ' . number_format((memory_get_usage() - $GLOBALS['memory_usage_start']) / 1024 / 1024, 2) . ' MB' . PHP_EOL);
+        unset($GLOBALS['memory_usage_start']);
     }
 
     /**
--- a/Classes/Core/BaseTestCase.php
+++ b/Classes/Core/BaseTestCase.php
@@ -40,9 +40,13 @@ abstract class BaseTestCase extends TestCase
 
     public static function setUpBeforeClass():void
     {
-        gc_collect_cycles();
-        $GLOBALS['memory_usage_start'] = memory_get_usage();
-        echo(PHP_EOL . '# ' . static::class . PHP_EOL);
+        if(getenv('TYPO3TESTINGFRAMEWORK_MEMORY_INFO')) {
+            echo PHP_EOL . '# \\' . static::class . PHP_EOL;
+
+            // Force garbage collection to measure memory used by these test-cases only.
+            gc_collect_cycles();
+            $GLOBALS['memory_usage_start'] = memory_get_usage();
+        }
 
         parent::setUpBeforeClass();
     }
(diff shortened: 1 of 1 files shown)


oliverklee:
Note: PHPUnit has some configuration concerning garbage collection which we might want to use:

https://docs.phpunit.de/en/11.1/configuration.html#the-controlgarbagecollector-attribute
https://docs.phpunit.de/en/11.1/configuration.html#the-numberoftestsbeforegarbagecollection-attribute

sbuerk:
Additionally to the phpunit garbage collection for the memory detection, this plugin may be usefull (not sure if there are others):

https://github.com/Roave/no-leaks

or could be used to improve and use proper phpunit 10 events to do these stuff without disturbing the output. For example, if people uses plugings to change the output this may wildy destroy the output even more then with the simple point output used in the core.

That would at least allow us to reduce the implementation in TF to a bare minimum, if we decide to keep the testcase property reflection detection in some way.

rr-it:
`<phpinfo controlGarbageCollector="true"/>` only runs garbage collection on every N-th test that is finished.

We need garbage collection to always run after all test methods of a test-case have finished. This would have maximum impact on memory usage.
See https://github.com/sebastianbergmann/phpunit/issues/5843

----

Additionally there is an opinion that we should not need to configure `<phpinfo controlGarbageCollector="true"/>`  at all - but PHPUnit should correctly handle memory cleanup by itself as default.

See https://review.typo3.org/c/Packages/TYPO3.CMS/+/84376/comment/546482c4_9aeb0d3e/
> i'm not convinced TF should deal with gc. this is a general phpunit issue that should be solved on this level if really needed.

lolli42:
Oliver, that's great to know!
If I'd be phpunit maintainer, I'd probably have implemented it exactly the same way, I came to essentially the same solution phpunit obviously has already: Trigger gc once in a while!
I don't agree with Ayke gc should be run after each test case: gc() needs to trade runtime with memory. memory is not top-1, we don't need to force "lowest memory usage possible" at all costs. We want a good trade-off instead. So doing it every x tests is imho a really good idea, and we should play around with these settings!

rr-it:
We have 518 functional test-cases with 7862 tests total.

- Run GC after each test-case: 
  GC runs 518 times.
- Run GC every 100 tests (PHPUnit default if GC handling is activated via `<phpinfo controlGarbageCollector="true"/>`)
  GC runs 78 times.

lolli42:
Hey. Recent phpunit 10 and 11 versions brought a series of changes, with one throwing away test case objects after test execution. This imho obsoletes this patch. I hope it's ok to close here.

PR #563, opened 2024-06-11
[TASK] Happify phpstan

lolli42:
Remove two obsolete exceptions - empty string will throw in \ReflectionClass by PHP.


Classes/Core/BaseTestCase.php
```diff
@@ -117,10 +117,6 @@ protected function getAccessibleMock(
         bool $callOriginalClone = true,
         bool $callAutoload = true
     ) {
-        if ($originalClassName === '') {
```
oliverklee, line 120: We probably can also remove the `@throws` annotation for both methods.

sbuerk (approved):

Issue #566, opened 2024-06-21
Functional tests fail on TYPO3v12: missing "install.php"


mbrodala:
### What are you trying to achieve?

Run functional tests on TYPO3v12.

### What do you get instead?

A failure since [version 8.1.0](https://github.com/TYPO3/testing-framework/releases/tag/8.1.0) of the testing framework:

https://github.com/pagemachine/typo3-formlog/actions/runs/9609757403/job/26504856080

> UnexpectedValueException: Source file (/.../typo3temp/var/tests/functional-3b7c664/typo3/sysext/install/Resources/Private/Php/install.php) was not found.

### How to reproduce the issue?

1. Run functional tests with `typo3/cms-*` in version `^12` without explicit `typo3/cms-install` and `typo3/testing-framework` 8.1.0
2. See the error

### Additional information you would like to provide?

This is a regression introduced with https://github.com/TYPO3/testing-framework/pull/543

### Specify some data of the environment

- TYPO3 testing framework version: 8.1.0
- TYPO3 version: 12.4.16
- TYPO3 installation type: composer / testing
- PHP version: 8

PR #567, opened 2024-06-21
[BUGFIX] Check for install.php existance

lolli42:
TF 8.1.0 removed 'composer req' of typo3/cms-install. Extension functional tests may now fail when trying to create the 'install.php' link - the file no longer exists if the extension itself has no requirement to typo3/cms-install, which most extensions don't have.

Resolves: #566


mbrodala (approved):
Works fine again now. :+1:

sbuerk (approved):

PR #620, opened 2024-09-06
[FEATURE] Support custom one-to-many relations inside fixtures

### LINKED ISSUE #619: Support custom one to many relations in yaml fixtures by nesting `entities` key
### Does your feature request relate to a problem?

Currently when writing yaml fixtures for test cases it is possible to model a datastructure by using the `entities` key. However the relation only correctly resolves when depending on the default `pid` column (e.g. content elements inside pages, see "example 1").
### Example 1
```yaml
entitySettings:
    '*':
        nodeColumnName: 'pid'
        columnNames: {id: 'uid', language: 'sys_language_uid'}
        defaultValues: {pid: 0}
    page:
        isNode: true
        tableName: 'pages'
        languageColumnNames: ['l10n_parent', 'l10n_source']
        defaultValues: {hidden: 0, doktype: 1}
    content:
        tableName: 'tt_content'
        languageColumnNames: ['l18n_parent', 'l10n_source']

entities:
    page:
        - self: {pid: 1, title: 'Sample Page'}
          entities:
              content:
                  - self: {CType: 'textmedia', header: 'Sample Content'}
```

Custom data structures with their own arbitrary relations cannot be modeled. Besides by creating them by their own and writing their incremented Uids (see `inline_record: 10000` at "example 2") by hand into the relation column which gets very messy when handling many. 
### Example 2
```yaml
__variables:
    - &pageStandard 1
    - &siteId 1

entitySettings:
    '*':
        nodeColumnName: 'pid'
        defaultValues: { pid: 0 }
    page:
        isNode: true
        tableName: 'pages'
        parentColumnName: 'pid'
        languageColumnNames: [ 'l10n_parent', 'l10n_source' ]
        columnNames: { type: 'doktype' }
        defaultValues: { hidden: 0, doktype: *pageStandard }
    productDetail:
        isNode: true
        tableName: 'tx_extname_domain_model_productdetail'
        languageColumnNames: [ 'l18n_parent', 'l10n_source' ]
        defaultValues:
            hidden: 0
    productPackageSize:
        tableName: 'tx_extname_domain_model_packagesize'
        languageColumnNames: [ 'l18n_parent', 'l10n_source' ]
        defaultValues:
            hidden: 0

entities:
    page:
        -   self: { pid: *siteId, title: 'Product Page A', type: 103, backendLayout: 'pagets__productPage', inline_record: 10000}
            entities:
                productDetail:
                    -   self: { title: 'Product A' }
                        entities:
                            productPackageSize:
                                -   self:
                                        title: 'Size A'
                                        size: 1
                                -   self:
                                        title: 'Size B'
                                        size: 2

```
Now lets assume one product may have many packageSizes (like also shown in "example 2"). And the TCA for the relation is configured as followed:
#### tx_extname_domain_model_productdetail
```php
[...]
'package_sizes' => [
    'label' => 'Package Sizes',
    'config' => [
        'type' => 'inline',
        'foreign_table' => 'tx_extname_domain_model_packagesize',
        'foreign_field' => 'package_size',
        'minitems' => 0,
        'maxitems' => 10,
    ],
],
[...]
```
#### tx_extname_domain_model_packagesize
```php
[...]
'product_detail' => [
    'config' => [
        'type' => 'passthrough',
    ],
],
[...]
```
So each `packageSize` has the `uid` of its `productDetail` set in its column `product_detail`.

Now to the root cause: The `TYPO3\CMS\Core\DataHandling\DataHandler` needs these kind of relations to be resolved by having the relation set the other way around. Meaning the datastructure the `DataHandler` expects needs to be like this:
```php
$data = [
    [...]
    'tx_extname_domain_model_productdetail' => [
        'NEW123456789' => [
            [...]
            'title' => 'Product A',
            'package_sizes' => 'NEW123,NEW321'
        ]
    ],
    'tx_extname_domain_model_packagesize' => [
        'NEW123' => [
            [...]
            'title' => 'Size A',
            'size' => 1
        ],
        'NEW321' => [
            [...]
            'title' => 'Size B',
            'size' => 2
        ]
    ]
];
```
Where the relation is set on the parent record and not like how it is stored in reality: on the child record.

### Preferred Solution
Being able to configure those kind of relations by setting some sort of "parent relation column"  in addition to the "pid" inside of the `entitySettings` of each fixture yaml.

nmaier95:
Summary:
Implement support to model custom relations of one-to-many relations inside fixture yamls by adding a new config field inside `entitySettings` for each table called `parentRelationColumnName`. When set, the data-array structure passed to the core `DataHandler` receives the relations like expected. The field needs to be set on a child record with the name of the parent record column name of the relation as its value.

Resolves #619

### DIFF
--- a/Classes/Core/Functional/Framework/DataHandling/Scenario/DataHandlerFactory.php
+++ b/Classes/Core/Functional/Framework/DataHandling/Scenario/DataHandlerFactory.php
@@ -96,7 +96,8 @@ class DataHandlerFactory
     private function processEntities(
         array $settings,
         ?string $nodeId = null,
-        ?string $parentId = null
+        ?string $parentId = null,
+        ?EntityConfiguration $parentEntityConfiguration = null,
     ): void {
         foreach ($settings as $entityName => $entitySettings) {
             $entityConfiguration = $this->provideEntityConfiguration($entityName);
@@ -105,7 +106,8 @@ class DataHandlerFactory
                     $entityConfiguration,
                     $itemSettings,
                     $nodeId,
-                    $parentId
+                    $parentId,
+                    $parentEntityConfiguration
                 );
             }
         }
@@ -115,7 +117,8 @@ class DataHandlerFactory
         EntityConfiguration $entityConfiguration,
         array $itemSettings,
         ?string $nodeId = null,
-        ?string $parentId = null
+        ?string $parentId = null,
+        ?EntityConfiguration $parentEntityConfiguration = null,
     ): void {
         $values = $this->processEntityValues(
             $entityConfiguration,
@@ -158,8 +161,22 @@ class DataHandlerFactory
             $this->processEntities(
                 $itemSettings['entities'],
                 $newId,
-                $parentId
+                $parentId,
+                $entityConfiguration,
             );
+        }
+        if (
+            $parentEntityConfiguration instanceof EntityConfiguration &&
+            $entityConfiguration->getParentRelationColumnName() !== null &&
+            $entityConfiguration->getParentRelationColumnName() !== '') {
+            $existingValues = array_filter(
+                explode(
+                    ',',
+                    (string)($this->dataMapPerWorkspace[$workspaceId][$parentEntityConfiguration->getTableName()][$nodeId][$entityConfiguration->getParentRelationColumnName()] ?? '')
+                )
+            );
+            $existingValues[] = $newId;
+            $this->dataMapPerWorkspace[$workspaceId][$parentEntityConfiguration->getTableName()][$nodeId][$entityConfiguration->getParentRelationColumnName()] = implode(',', $existingValues);
         }
     }
 
--- a/Classes/Core/Functional/Framework/DataHandling/Scenario/EntityConfiguration.php
+++ b/Classes/Core/Functional/Framework/DataHandling/Scenario/EntityConfiguration.php
@@ -25,6 +25,7 @@ class EntityConfiguration
     private bool $isNode = false;
     private ?string $tableName = null;
     private ?string $parentColumnName = null;
+    private ?string $parentRelationColumnName = null;
     private ?string $nodeColumnName = null;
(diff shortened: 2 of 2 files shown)


Classes/Core/Functional/Framework/DataHandling/Scenario/EntityConfiguration.php
```diff
@@ -86,6 +90,11 @@ public function getParentColumnName(): ?string
         return $this->parentColumnName;
     }
 
+    public function getParentRelationColumnName(): ?string
+    {
+        return $this->parentRelationColumnName;
+    }
```
sascha-egerer, line 96: Setting a parent relation column name does not make sense here as a table can be used in different tables and fields. So the relation is defined by the parent and not by the entity itself.

Classes/Core/Functional/Framework/DataHandling/Scenario/DataHandlerFactory.php
```diff
@@ -158,8 +161,22 @@ private function processEntityItem(
...
+            $entityConfiguration->getParentRelationColumnName() !== null &&
+            $entityConfiguration->getParentRelationColumnName() !== '') {
+            $existingValues = array_filter(
+                explode(
+                    ',',
+                    (string)($this->dataMapPerWorkspace[$workspaceId][$parentEntityConfiguration->getTableName()][$nodeId][$entityConfiguration->getParentRelationColumnName()] ?? '')
+                )
             );
+            $existingValues[] = $newId;
+            $this->dataMapPerWorkspace[$workspaceId][$parentEntityConfiguration->getTableName()][$nodeId][$entityConfiguration->getParentRelationColumnName()] = implode(',', $existingValues);
```
sascha-egerer, line 179: This implementation would only work for entities that are defined as “nodes”. It would therefore only work for pages.

nmaier95:
closed because it became obsolete.

PR #625, opened 2024-10-08
[BUGFIX] Respect overridden properties in AbstractPageTree

lolli42:
This changes all property accesses within AbstractPageTree to use static instead of self. Since the class is abstract and therefore subject to class inheritance, its properties may also change in subclasses. Using self to access them would hide potentially overridden properties (late static binding).


sbuerk (approved):

lolli42:
This is one part of https://github.com/TYPO3/testing-framework/pull/479 - core main stays green with this.

PR #626, opened 2024-10-08
[BUGFIX] Harden XPath query to limit search for page tree node

lolli42:
Even if page tree nodes are queried from a given context node, the used XPath query searches for all elements in the document because of the `//` syntax. This may lead to possible test failures in case the queried node text appears multiple times in the document.

In order to avoid such behaviors and actually limit the XPath query to the given context, it is now prefixed with a dot (`.`). In addition, only following siblings of the current node are included since page tree nodes are presented as flat node structure. This way, the current context is now actually taken into account.


lolli42:
Part of https://github.com/TYPO3/testing-framework/pull/479

sbuerk (approved):

Issue #577, opened 2024-06-24
Question about normalizedParams in frontend sub-requests


nhovratov:
Not sure if this is the correct place to open the issue, however I have a problem migrating Functional Tests from v12 to v13 with the latest version of the testing framework:

I have a test, where I am checking for asset paths in the output HTML. The problem now is, since v13 the paths aren't prefixed anymore. See failing run: https://github.com/nhovratov/content-blocks/actions/runs/9550659756/job/26323141221

While digging deeper, it looks like the property "sitePath" from normalizedParams is different in both versions. This property is used when absRefPrefix is set to "auto", which is always the case in sub requests.

Here some debug outputs:

v12 (sitePath: "/")
![Bildschirmfoto von 2024-06-19 18-42-55](https://github.com/TYPO3/testing-framework/assets/19343425/945a1b32-4dd0-4a1d-ae5f-41d257369927)

v13 (sitePath: ""), siteScript "ttp" also seems cut off.
![Bildschirmfoto 2024-06-19 um 18 38 26](https://github.com/TYPO3/testing-framework/assets/19343425/cbb51d15-0120-49e2-8fae-4c17c6fd0e54)

lolli42:
Hey.

I'm afraid we need some more information here to understand what to do: This smells like a bug in either TF or core, but we're unsure.

Is this still an issue with current v13 main/9 TF?
What do you think where this stems from?
Is there a core functional test that shows this behavior as well? How to reproduce / debug best?

nhovratov:
Here is a functional test showcasing the error with TYPO3/main, TF/main

https://github.com/nhovratov/content-blocks/actions/runs/11259518816/job/31308673475

The exact (!) same setup will work when run from Core monorepo. So it looks like there is some difference.

sbuerk:
I had already a chat with @nhovratov. I know what is involved,
and already did some invetigations - having multiple "solutions"
now available which I need to investigate / analyze against/further:

* core monorepo structure
* usual usage for single extension testing (with changed install root -> ./Build/Web for example)
* usual usage for project/custom monorepo testing

PR #630, opened 2024-10-13
[BUGFIX] Ensure correct path calculation on system build

sbuerk:
TYPO3 core `SystemEnvironmentBuilder` has been implement
normal TYPO3 usages in mind respecting possible instance
setup scenarios, to build low level system environment
before taking configuration into account.

Path calculations are based on different indicators,
for example if TYPO3 is used in composer mode based
on the PHP define() TYPO3_COMPOSER_MODE set during
by the `typo3/cms-composer-installers` composer plugin
during `composer install` actions.

Per nature, PHP defines are immutable and cannot be
changed anymore during runtime.

As `typo3/testing-framework` is only installabe using
composer the TYPO3_COMPOSER_MODE define is always set
but is invalid for functional test instances build in
legacy mode and making invalid path calculations.
That was hidden quite some time now, and is related to
a chain of changes over a very long period of time and
finally popped up in early TYPO3 v13 development after
introducing the configurable backend url feature.

Functional tests building relative url for resources
or links could run into the issue not retrieving the
leading slash anymore due to path calculation errrors,
leading to follup issues in `NormalizedParams` create
method when using frontend requests.

Eventually more developers run into that issue, but
simply adjusted the test expectation instead analyzing
the root of the curse and thus not reported it, which
has been done recenently.

To fix this issue changes on two fronts have to be made,
in the TYPO3 core `SystemEnvironmentBuilder` [1] and
in the testing-framework implementation.

Note that only both changes together solves this issue,
as the testing-framework change alone would not be
executed due to using `self` for static method calls
in the TYPO3 implementation.

[1] https://review.typo3.org/c/Packages/TYPO3.CMS/+/86569

Resolves: #577
Releases: main, 8

### DIFF
--- a/Classes/Core/SystemEnvironmentBuilder.php
+++ b/Classes/Core/SystemEnvironmentBuilder.php
@@ -40,13 +40,16 @@ use TYPO3\CMS\Core\Core\SystemEnvironmentBuilder as CoreSystemEnvironmentBuilder
  */
 class SystemEnvironmentBuilder extends CoreSystemEnvironmentBuilder
 {
-    public static function run(int $entryPointLevel = 0, int $requestType = 0, bool $composerMode = false): void
+    private static ?bool $composerMode = null;
+
+    public static function run(int $entryPointLevel = 0, int $requestType = 0, ?bool $composerMode = null): void
     {
-        CoreSystemEnvironmentBuilder::run($entryPointLevel, $requestType);
+        self::$composerMode = $composerMode;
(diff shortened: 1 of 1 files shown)


lolli42:
Note: We are aware this is a hack. This change follows the existing current strategy of overwriting core SystemEnvironmentBuilder and extends it slightly. We'll eat this technical debt for now.

lolli42 (approved):

PR #669, opened 2025-04-10
[BUGFIX] Allow to process versions of language variants

bmack:
This change allows for YAML-based fixtures
to also add version variants for a language variant thus, making it fully overlayable.

See https://review.typo3.org/c/Packages/TYPO3.CMS/+/72015


sbuerk:
Looks good, testing it agaist the core change first.

sbuerk (approved):

PR #703, opened 2026-02-20
[TASK] Enable strict mode for acceptance tests

bmack:


oliverklee (approved):

sbuerk:
@bmack Code wise okay, not checked it yet.

Questions:

* core-main -> green with nightly ? (gerrit change?/pipelines?)
* core-13 -> green with nightly ? (gerrit change?/pipelines?)

If that's green, I guess we should also look into the functional tests and doing the same. If functionals are not green, are we save to enable it for acceptance tests than ?

lolli42:
Hey.

I hope it is ok to close here: We're about to remove the codeception related files from TF main. For TF v9 they are marked deprecated to retain core v13&v14 compat, but I think the patch is not crucially important and would affect old stable core v13 only. A such, I hope it is ok to close the PR since I don't really want to fiddle with the v13 implications of this change anymore.

Let's re-push in case this is still important to reconsider.

PR #709, opened 2026-03-15
[TASK] Make getInstanceIdentifier() and getInstancePath() non-static for overridability

dkd-kaehm:
Convert protected static methods to protected instance methods and mark as @internal.
This allows subclasses to override these methods for proper test isolation in parallel
execution (Paratest), without making them part of the public API.

Instead of using static:: binding, use $this-> for natural method overriding.
Solves the need for test isolation in single-test-level parallelization while
signaling that these are internal implementation details.

Fixes: #708

### DIFF
--- a/Classes/Core/Functional/FunctionalTestCase.php
+++ b/Classes/Core/Functional/FunctionalTestCase.php
@@ -284,8 +284,8 @@ abstract class FunctionalTestCase extends BaseTestCase implements ContainerInter
             self::markTestSkipped('Functional tests must be called through phpunit on CLI');
         }
 
-        $this->identifier = self::getInstanceIdentifier();
-        $this->instancePath = self::getInstancePath();
+        $this->identifier = static::getInstanceIdentifier();
+        $this->instancePath = static::getInstancePath();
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
@lolli42 
switched to non-static.

lolli42:
Hey Rafael. Sorry for the delay. I had to first fix CI today, and had another look at the change now.
I rebased it and gave it some thoughts.
I'm sorry for discussing so many details on this rather tiny change, but reading it again I'm now actually a bit unsure if people may hurt us by making especially getInstancePath() non-static, this might be considered breaking in seldom cases (and core probably actually stumbles here). I'll try to pull @sbuerk into this to come up with a final conclusion to solve your initial issue, soon. Please be patient, I'm sure we'll find consensus.

lolli42:
Ok. Had a chat with Stefan on this. We will go with your initial approach (without the test). This is not breaking an risk less. Sorry for the confusion and thanks for your work.

sbuerk (approved):

sbuerk (approved):

lolli42 (approved):

PR #715, opened 2026-04-22
[TASK] Allow TYPO3 v15-dev

bmack:


bmack:
won't work. I assume we need to fork first.

PR #726, opened 2026-07-21
[TASK] Allow installation of guzzlehttp/psr7 v3

bnf:
In order to prepare for guzzlehttp/guzzle v8 in
TYPO3 core, we allow the installation of the
`guzzlehttp/psr7` helper package.

composer req guzzlehttp/psr7:"^2.5.0 || ^3.0"

Releases: main, 9, 8

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
   "require": {
     "php": "^8.2",
     "composer/class-map-generator": "^1.3.4",
-    "guzzlehttp/psr7": "^2.5.0",
+    "guzzlehttp/psr7": "^2.5.0 || ^3.0",
     "phpunit/phpunit": "^11.2.5 || ^12.1.2 || ^13.0.2",
     "psr/container": "^2.0",
     "typo3/cms-backend": "14.*.*@dev || 15.*.*@dev",


sbuerk:
Hey @bnf , do we really need to go down to `6` ? 

In general 6 & 7 is not really considered `maintained` anymore and there could be now a some issue (regarding pipelines) and similar. Not that I'm trying to block that, but we should talk about that because I need to ensure that everything works pipeline wise so I can do than all the releases.

I checked the `composer.json` down to `6` for the current require constraint and tried to elaborate what would be the constraint to set for each version (backport wise).

Can you check and confirm it ?

| branch | current          | new (suggested)              |
|--------|------------------|------------------------------|
| main   | `^2.5.0`         | `^2.5.0 \|\| ^3.0`           |
| 9      | `^2.5.0`         | `^2.5.0 \|\| ^3.0`           |
| 8      | `^2.5.0`         | `^2.5.0 \|\| ^3.0`           |
| 7      | `^2.5.0`         | `^2.5.0 \|\| ^3.0`           |
| 6      | `^1.7 \|\| ^2.0` | `^1.7 \|\| ^2.5.0 \|\| ^3.0` |

From what I checked on packagist.org so far that should be doable / safe.

bnf:
Hey @sbuerk, let's go with 9 and 8 for now.
I'll evaluate if we need v7 and v6 as well. Thanks!

sbuerk (approved):

PR #748, opened 2026-09-16
[TASK] Speed up functional data set imports

sbuerk:
Backport of #743 to branch `9`.

Clean cherry-pick of db2ff35 from `main`, no adaptations needed. Only the
commit message was rewrapped to 72 chars and got the
`Releases: main, 10, 9` footer.

### TYPO3 v13 compatibility

All core APIs used exist with the same signatures in TYPO3 v13.4, verified
against the `v13.4.35` sources of `typo3/cms-core`:

* `Connection::getSchemaInformation()` and
  `SchemaInformation::listTableColumnInfos(string $tableName): array<string, ColumnInfo>`,
  keyed by the untouched column name (`SchemaInformation::buildTableInformation()`).
* `TYPO3\CMS\Core\Database\Schema\Information\ColumnInfo` is a
  `final readonly` class with public `string $name`, `bool $autoincrement`
  and `getType(): Type`.
* The PostgreSQL sequence reset only uses plain SQL
  (`SETVAL()` / `PG_GET_SERIAL_SEQUENCE()`), no core API.

### Quality gates

Run locally with `Build/Scripts/runTests.sh`, PHP 8.2 (composer resolved
`typo3/cms-core` 14.3.x-dev):

* `-s composerUpdate`: passed
* `-s cgl -n`: passed
* `-s lint`: passed
* `-s phpstan`: passed
* `-s unit`: passed

### Original description

CSV data set imports currently perform full Doctrine schema introspection for every imported table. PostgreSQL additionally queries several system catalog tables before synchronizing each auto-increment sequence.

Use TYPO3's cached schema information to obtain column types and identify the auto-increment column. Skip sequence handling for tables without such a column.

Pass the known column name to the sequence reset helper and use PG_GET_SERIAL_SEQUENCE() to synchronize PostgreSQL sequences in a single query. Retain the existing catalog lookup as a fallback for callers that do not provide column metadata.

In a complete PostgreSQL functional suite run with the same ten-way split as CI, aggregate runtime decreased from 2:26:36.598 to 2:14:04.534, saving 12:32.064 or 8.55%. The slowest chunk improved from 16:04.463 to 14:47.040, saving 8.03%. All 12,484 tests passed.


sbuerk:
Change already on branch 9 through another pr.

